PR #1, opened 2017-08-24
Initialize Lerna-managed monorepo containing htz-react-base

exogen:
This adds the `htz-react-base` package for encapsulating development tasks.

It takes inspiration from both Next.js and `react-scripts` (from `create-react-app`). You include it in your `devDependencies` and gain the `htz-scripts` binary.

The current working scripts:

* `htz-scripts build` – Transpiles a reusable library like `htz-components` so it can be imported via Node and webpack.
* `htz-scripts format` – Runs [Prettier](https://prettier.io/) (we can change the preferences whenever).
* `htz-scripts test` – Runs Jest, although there is currently some configuration issue.


packages/htz-react-base/package.json
```diff
@@ -0,0 +1,30 @@
+{
+  "name": "htz-react-base",
+  "version": "0.0.0",
+  "private": true,
+  "description": "",
+  "main": "index.js",
+  "bin": {
+    "htz-scripts": "./bin/htz-scripts.js"
+  },
+  "scripts": {
```
TxHawks: Might be just some local issue with my setup here, but with `yearn run <task-name>` I get the following error:
```sh
sh: ./bin/htz-scripts.js format: No such file or directory
```

Works just fine with `npm run <task-name>`.

Strangely, wrapping the command in quotes fixes it, so this does work:
```js
  "scripts": {
     "format": "\"./bin/htz-scripts.js format\"",
    "test": "\"./bin/htz-scripts.js test\""
},
```

TxHawks:
Other than the inline comments, I really do think we should consider handling top-level dependencies like React, Fela, Apollo, etc. here as well...

I can't think of a better solution, and I foresee an endless sea of duplications and conflicts if we don't.

README.md
```diff
@@ -1 +1,37 @@
...
+Use [Yarn](https://yarnpkg.com/) to install Lerna:
+
+```shell
+$ yarn
+```
+
+Now use Lerna to set up the managed packages:
+
+```shell
+$ lerna bootstrap
```
exogen: Good call, I've got `./node_modules/.bin` permanently in my `$PATH` so didn't notice. Will add a `script` for it.

packages/htz-react-base/package.json
TxHawks: This was an issue with my local setup. Not sure why, but Homebrew installed a very old version of yarn. Resolved itself when I upgraded

PR #2, opened 2017-08-24
Add htz-components package with working styleguide

exogen:
This package serves as a place to collect and develop components. ~~I'm not sure if the site-specific components should live here or in their respective apps, but they currently live here.~~

To see the components:
* `yarn run styleguide`

This also uses `htz-react-base`, so all those `htz-scripts` commands work, and they're also aliased in `package.json` for ease of use.


packages/htz-components/styleguide.config.js
```diff
@@ -0,0 +1,44 @@
+module.exports = {
+  title: 'Haaretz Components',
+  showCode: true,
+  showUsage: true,
+  sections: [
```
TxHawks: I think, for organization and scalability, each component should always be in its own directory, so maybe something more like:
```js
  {
    name: 'Shared Components',
    components: 'src/components/shared/**/*.js'
  },
  {
    name: 'Site Specific Components',
    sections: [
      {
        name: 'haaretz.co.il',
        components: 'src/components/haaretz.co.il/**/*.js'
      },
      {
        name: 'haaretz.com',
        components: 'src/components/haaretz.com/**/*.js'
      },
    ]
  },
```

And `src/components/shared` should look like
```
shared
|
|____ Article
|      |____ Article.js
|
|____ Button
       |____ Button.js
       |____ Button.md
       |____ Button.test.js
```

exogen: Sounds good to me. I like putting multiple closely feature related in the same directory as well, e.g. `Search/SearchInput.js`, `Search/SearchButton.js`, `Search/SearchIcon.js` – how do you feel about that, or would you prefer those be in their own directories as well?

I'm hoping we can also make things nicer to import for consumers, because needing to `import Button from 'htz-components/lib/components/Button/Button'` as opposed to `import { Button } from 'htz-components'` can get tedious – just want to make sure the import tree-shaking is working properly in Next, so the bundle size doesn't blow up.

TxHawks: I think a pattern like
```
src/components
|
|__ index.js
|__search/
   |__index.js
   |__SearchInput/
   |  |__SearchInput.js
   |  |__SearchInput.md
   |  |__SearchInput.test.js
   |
   |__SearchButton/
      |__SearchButton.js
      |__SearchButton.md
      |__SearchButton.test.js
```

with `index.js` importing all components in its dir allows the best of both worlds. It does put more responsibility on the dev to remember to add components to index, but maybe we can automate that somehow. Shouldn't be too hard

packages/htz-components/package.json
```diff
@@ -0,0 +1,35 @@
...
+    "format": "htz-scripts format",
+    "prepare": "npm run build",
+    "test": "htz-scripts test",
+    "styleguide": "styleguidist server",
+    "styleguide:build": "styleguidist build"
+  },
+  "peerDependencies": {
+    "fela": "^5.1.1",
+    "next": "^3.0.6",
+    "react": "^15.6.1",
```
TxHawks: This should probably be: `"react": "^15.6.1 || ^16.0.0-0",` to make it easier to play with `16` and eventually switch to it when it's ready

packages/htz-components/src/components/Button/Button.js
```diff
@@ -0,0 +1,9 @@
+import React from 'react'
+
+export default ({ children, ...props }) => {
```
TxHawks: I love store functions, but prefer named functions for stateless components. It is useful for components to have a name and debugging is easier when function names are not inferred.

This pattern can end up showing the component as `<StatelessComponent>` in the React devtools

packages/htz-components/src/index.js
```diff
@@ -0,0 +1,4 @@
+export * from './components'
+export { default as styleRenderer } from './style/renderer'
+export { default as StyleProvider } from './style/StyleProvider'
+export { default as HaaretzDocument } from './document'
```
TxHawks: See previous comment about explicit `import`s

packages/htz-react-base/styleguide.config.js
```diff
@@ -0,0 +1,76 @@
...
+    process.cwd(),
+    'htz-components/src/components/StyleProvider/StyleProvider'
+  ) ||
+  resolveFrom.silent(
+    process.cwd(),
+    'htz-components/dist/lib/components/StyleProvider/StyleProvider'
+  )
+
+module.exports = {
+  title: 'Haaretz Components',
```
TxHawks: Shouldn't this be set on a per-styleguide base, or is this just an overwritable default?

exogen: Yes, just a default (see #3 where the app sets it). Could make this an even more generic default and just be "Components" if that's better.

packages/htz-components/src/components/index.js
```diff
@@ -0,0 +1,2 @@
+export { default as Article } from './Article/Article'
+export { default as Button } from './Button/Button'
```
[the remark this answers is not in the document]
exogen: 👍  I actually feel the same way, this always trips me up...why didn't they just make `export Foo from 'foo'` work?

packages/htz-components/src/index.js
exogen: This comment also made me notice these exports are wrong since I moved `StyleProvider` into `components`, will fix.

packages/htz-components/src/components/index.js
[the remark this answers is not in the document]
exogen: How do you feel about killing this `components/index.js` and just exporting them in the root `index.js`? `import { Article } from 'htz-components'` is going to be nicer than `import { Article } from 'htz-components/dist/lib/components'`.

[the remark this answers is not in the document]
TxHawks: Was just about to suggest this :)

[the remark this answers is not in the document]
TxHawks: But we do need to make sure tree-shaking works. Should I open an issue so that we do t forget about it?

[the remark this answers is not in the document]
exogen: @TxHawks Sure, thanks!

packages/htz-components/.babelrc
```diff
@@ -0,0 +1,3 @@
+{
+  "presets": ["htz-react-base/babel"]
+}
```
exogen: I didn't want to have to add this to every component module we make, but it's necessary for Jest to know how to transform test files at the moment.

exogen:
@TxHawks What are the most important changes I should make here before this is ready?

For the styleguide component path / import line issue, should we just assume that any component in a module can be imported from the root e.g. `import { Button } from 'htz-components'`? I fear it would be too much code to determine where the expected/public import path is otherwise.

TxHawks:
I'm not sure I understand the the import path question? Why do we need to make assumptions? isn't that the developers' responsibility? I mean, the way it is set up now, developers need to add importable components to `src/index.js`. Do you want to automate it?

packages/htz-components/package.json
```diff
@@ -0,0 +1,38 @@
...
+  "peerDependencies": {
+    "fela": "^5.1.1",
+    "next": "^3.0.6",
+    "react": "^15.6.1 || ^16.0.0-0",
+    "react-fela": "^5.2.1"
+  },
+  "dependencies": {
+    "fela-dom": "^5.0.7",
+    "fela-preset-web": "^5.0.7",
+    "prop-types": "^15.5.10"
```
TxHawks: Why are all these `dependencies` and not `devDependencies` and `peerDependencies`. Surly consumer apps will need them as well

exogen:
Per [this comment](https://github.com/Haaretz/htz-frontend/pull/2#discussion_r135381501), it's the styleguide docs that need to make assumptions. 

In order to render this line as an `import` statement:
screen shot 2017-08-28 at 12 32 14 am

We only know the source (pre-transpiled) source path at this point, and it's not easy to determine where that component is expected to be imported from by consumers. For example, even if you wanted to reach directly into that component's file, you wouldn't import from `src` but rather `dist/lib`. Instead of a bunch of file and AST parsing code, I think we should just assume that any component can be imported from the package's `main` module.

TxHawks:
Oh, okay, I just didn't understand the context. I'm with you on this.

The only other issue left is one I notice just now, about using `dependencies` vs `devDependencies` and `peerDependencies` [here](https://github.com/Haaretz/htz-frontend/pull/2/files/2ed7e6924cb1de80c7aa5d24d548ba8d2430b6f3#diff-5fe826decfbd2d747a6ef68987af2ef9R24)

Otherwise, LGTM  :rocket:

packages/htz-components/package.json
exogen: When I first learned that they were splitting out `prop-types` I also had this question and expected it to be `peerDependencies`, but it's the official recommendation: https://www.npmjs.com/package/prop-types#how-to-depend-on-this-package

For `fela-dom` and `fela-preset-web`, I actually don't think any other module will need them, because `StyleProvider` encapsulates them. The `StyleProvider` component is essentially the public interface to Fela we expose to any other package. Can follow up in the morning with more details/look into it more.

TxHawks: No, you're right. They do it the [same way](https://github.com/cloudflare/cf-ui/blob/master/packages/cf-style-provider/package.json) at Cloudflare

TxHawks: We need to make sure `prop-types` is excluded from the build though:
> For UMD bundles of your comoponents, make sure you don’t include PropTypes in the build. Usually this is done by marking it as an external (the specifics depend on your bundler), just like you do with React.

TxHawks: Should I open an issue, or is it already taken care of?

exogen: I would expect Next's webpack config to take care of this, but looking through it, I don't see this in there – open an issue so we can look into it.

TxHawks:
Is there currently an env variable to determine if we are in `dev` or `production` mode? I'd probably want, for instance, to load some Fela plugins like [fela-plugin-validator](https://github.com/rofrischmann/fela/tree/master/packages/fela-plugin-validator) or only when developing and keep them out of production

exogen:
We'll use the conventional `NODE_ENV`, but I don't think anything except Next is taking it into account yet. I'll update this PR with that plugin in the morning.

exogen:
@TxHawks Made all outstanding changes, should be good to go!

PR #3, opened 2017-08-24
Add site application packages using Next.js and mock API

exogen:
[This PR uses feat/add-htz-components as its base branch; when that is merged, we should point it at master.]

This app uses both `htz-react-base` and `htz-components`.

You can run it like so:
* `yarn run dev`

It uses a small mocked "story" API to simulate pages fetching the article data that they need.


TxHawks:
I know this is really just a demo, but we should still probably start with `haaretz.co.il` rather than `haaretz.com`. That's our main website, and the one we will build first.

Our internal naming is:
htz: haarez.co.il
hdc: haaretz.com
tm: themarker.com

exogen:
@TxHawks Do you think I should add an app for themarker.com to this PR, or maybe remove the haaretz.com app? Especially while things are getting set up, it might be more work to keep 2 or 3 apps in sync with all the changes coming down the line. If we just work on haaretz.co.il, that can serve as a reference for the others when the time is right.

TxHawks:
Yes, removing haaretz.com is totally fine

exogen:
Now that we're trying to inspect production app bundles (e.g. for #12, #6), I think we should get this into `master` as soon as possible, otherwise it's much harder to run other branches like #28 on it.

The rendered site and mock API are obviously placeholders, but we need something here.

@TxHawks, ready to make whatever changes you deem necessary before merging.

TxHawks:
I think we can merge this for now. Seeing as it's really just a mock we can make changes later

PR #9, opened 2017-08-28
Add documentation and new update-scripts command for htz-react-base

exogen:
* Add installation and usage docs for `htz-react-base`.
* Add a new `update-scripts` command, for easily syncing `package.json` scripts with the commands available via `htz-scripts`.


packages/htz-react-base/README.md
```diff
@@ -0,0 +1,95 @@
...
+$ yarn add -D htz-react-base
+```
+
+## Usage
+
+Installing `htz-react-base` will make an `htz-scripts` executable available to
+the consuming package, which can be run like so:
+
+```console
+$ ./node_modules/.bin/htz-scripts <command>
```
TxHawks: npm >= 5 provides nicer ergonomics for this with the `[npx](https://www.npmjs.com/package/npx)` command, which turns this to the much more convenient `npx htz-scripts <command>`.

packages/htz-react-base/README.md
```diff
@@ -0,0 +1,95 @@
...
+
+Build production distributable files. For apps, this will involve running the
+[Next.js](https://github.com/zeit/next.js) build process. For modules, this will
+usually involve running Babel and/or webpack.
+
+### format
+
+Format code in the current directory with [Prettier](https://prettier.io/).
+You should do this before opening any pull request that contains code, for
+consistency.
```
TxHawks: We automate this with a git pre-commit with [Husky](https://www.npmjs.com/package/husky). I'll open an issue so that I don't forget to do this.

packages/htz-react-base/README.md
```diff
@@ -0,0 +1,95 @@
...
+The output will be written to `dist/styleguide`.
+
+### test
+
+Run the package’s test suite with [Jest](https://facebook.github.io/jest/).
+
+### update-scripts
+
+Update the `scripts` field in `package.json` to contain aliases for `htz-scripts`
+commands. Use `--overwrite` to clobber any existing scripts with the same name.
```
TxHawks: We should consider making `--overwrite` the default, since I think it will be the more common use-case and we probably want to maintain consistency between the different packages when updating base scripts.

We can add an explicit `--no-overwrite` flag for when this behavior is really needed. Both options can produce unexpected result, but I feel that with everything being under source control and revertable, this is the safer and more straightforward workflow

packages/htz-react-base/scripts/update-scripts.js
```diff
@@ -0,0 +1,63 @@
...
+  styleguide: 'htz-scripts styleguide',
+  'styleguide:build': 'htz-scripts styleguide:build'
+}
+
+const packagePath = resolveFrom(process.cwd(), './package.json')
+const packageInfo = require(packagePath)
+
+// If --overwrite is specified, overwrite any existing scripts having the same
+// name with the version here.
+const overwrite = process.argv.indexOf('--overwrite') >= 2
```
TxHawks: See comment above about `--overwrite` being default behavior

packages/htz-react-base/README.md
exogen: Agreed 👍

packages/htz-react-base/README.md
exogen: Sounds good to me, I tend to make the safest option the default but I think you're right that we'll want them to stay in sync anyway. When it overwrites a script it prints what it used to be, so it's easy to fix even without git

PR #28, opened 2017-08-30
Build package.json:main fields for different targets

### LINKED ISSUE #4: Build related fields in packages' `package.json`
Once we get the build scripts nailed down, we should add
```json
  "main": "./dist/lib/index.js",
  "module":  "./dist/module/index.js",
  "esnext:main": "./dist/es/index.js",
```
to all relevant `pacckage.json` files

Pre-transpile JSX, but leaves other JavaScript syntax to be handled by the consuming app via webpack’s esnext field and babel-preset-env. [[1](http://2ality.com/2017/06/pkg-esnext.html)]  [[2](http://2ality.com/2017/07/npm-packages-via-babel.html)]

exogen:
This updates the `build` command to run several subtasks:

* Run Babel to generate `/dist/lib`, which targets Node environments with CommonJS modules.
* Run Babel to generate `/dist/esm`, which targets webpack with ECMAScript modules.
* Run Babel to generate `/dist/jsnext`, which applies minimal transformations so that consumers can do further processing (e.g. with their own `babel-preset-env` configuration).
* If the package is a Next.js app, run `next build` instead.

Fixes #4.


packages/htz-react-base/scripts/build/index.js
```diff
@@ -0,0 +1,78 @@
...
+  })
+}
+
+function buildApp() {
+  require('./next')
+}
+
+function buildLibrary() {
+  const envScripts = {
+    commonjs: require.resolve('./lib'),
```
TxHawks: Can we rename this to `UMD`?

packages/htz-react-base/scripts/build/index.js
```diff
@@ -0,0 +1,78 @@
...
+  if (isDir) {
+    console.log('Next.js app package detected.')
+    buildApp()
+  } else {
+    checkDir('src').then(isDir => {
+      if (isDir) {
+        console.log('Library package detected.')
+        buildLibrary()
+      } else {
+        console.error('Could not determine how to build the current package.')
```
TxHawks: Maybe we should allow passing an `--app` or `--lib` flag followed by a `path` argument, for cases we have a package with non-standard sir structure. Not crucial or urgent though.
Let's not waste any time on this now. I'll open an issue one this PR is merged to add this in the future myself

packages/htz-components/package.json
```diff
@@ -10,6 +10,8 @@
     "access": "public"
   },
   "main": "dist/lib/index.js",
+  "module": "dist/esm/index.js",
+  "jsnext:main": "dist/jsnext/index.js",
```
TxHawks: This should be `esnext:main`.
Also, will we need different builds for `main` and `browser`?

packages/htz-react-base/babel.js
```diff
@@ -1,30 +1,76 @@
...
+        modules: false
+      }
+    ])
+    break
+  // The `jsnext` environment is used to generate `/dist/jsnext` and will be
+  // used by advanced webpack setups that are specifically configured to
+  // handle syntax with as few transforms pre-applied as possible. Additional
+  // transpilation will be necessary (via `babel-loader`), but consuming apps
+  // can decide what to do for themselves.
+  case 'jsnext':
```
TxHawks: Should probably be `esnext` for consistency

packages/htz-react-base/scripts/build/jsnext.js
```diff
@@ -0,0 +1,9 @@
+process.env.BABEL_ENV = process.env.BABEL_ENV || 'jsnext'
```
TxHawks: Should probably be `esnext` for consistency

packages/htz-react-base/scripts/build/lib.js
```diff
@@ -0,0 +1,9 @@
+process.env.BABEL_ENV = process.env.BABEL_ENV || 'commonjs'
```
TxHawks: Use `UMD` instead of `commonjs`?

packages/htz-react-base/scripts/build/index.js
```diff
@@ -0,0 +1,78 @@
...
+
+function buildApp() {
+  require('./next')
+}
+
+function buildLibrary() {
+  const envScripts = {
+    commonjs: require.resolve('./lib'),
+    esm: require.resolve('./esm'),
+    jsnext: require.resolve('./jsnext')
```
TxHawks: Let's rename this to `esnext` for Consistency

packages/htz-react-base/scripts/build/next.js
```diff
@@ -0,0 +1,4 @@
+const resolveFrom = require('resolve-from')
+
+process.argv.push('build')
+require(resolveFrom(process.cwd(), 'next/dist/bin/next'))
```
TxHawks: We should extend Next's webpack config (can we?) to consume and transpile `esnext` builds for packages whose `package.json` has the `esnext:main` field

packages/htz-components/package.json
exogen: I deliberately named it something other than `esnext:main`, because `esnext:main` was the alternative name for `module` before a final one was decided upon, so this risks consumers using that webpack configuration accidentally getting untranspiled code.

I thought that 2ality article used `jsnext:main`, but it's actually just plain `esnext`. Should we use that? Or are you sure about `esnext:main`?

packages/htz-react-base/babel.js
```diff
@@ -1,30 +1,76 @@
...
+// `babel-preset-env` until such transforms are official.
+const presets = ['react']
+const plugins = ['transform-object-rest-spread', 'transform-class-properties']
+const ignore = []
+
+switch (process.env.BABEL_ENV) {
+  // The `commonjs` environment is used to generate `/dist/lib` and will be
+  // used for traditional Node imports, including during server side rendering
+  // (SSR). It should target the minimum supported Node.js version.
+  case 'commonjs':
```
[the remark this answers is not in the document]
exogen: It's not accurate though, UMD is something else – that's when you check for `typeof define`, etc. to support multiple environments, which isn't how modules are transpiled in this scenario.

packages/htz-react-base/babel.js
```diff
@@ -1,30 +1,76 @@
...
+    presets.push([
       'env',
       {
         targets: {
-          node: 'current'
+          node: '6.10'
```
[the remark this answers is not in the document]
exogen: Sounds good to me!

packages/htz-components/package.json
TxHawks: `esnext` is fine by me, unless we want seperate packages for SSR and browser, for example use less transpiled packages on the server because we only need to support node 8

packages/htz-react-base/babel.js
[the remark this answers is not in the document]
TxHawks: Shouldn't we be doing UMD so that we can use this in the browser too? Unless we need seperate packages for SSR and browser anyway (see comment above about `esnext` field)

packages/htz-react-base/babel.js
```diff
@@ -1,30 +1,76 @@
...
+  // by bundlers like webpack 2+ with no extra configuration necessary.
+  case 'esm':
+    presets.push([
+      'env',
+      {
+        targets: {
+          // The browser list having any slimming effect on the selected
+          // transforms is mostly aspirational due to the UglifyJS requirement,
+          // see below.
+          browsers: require('./browsers'),
```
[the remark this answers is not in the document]
exogen: The impression I'm under is that Haaretz apps will use the `esnext` distribution per that 2ality post, in which case they would specify their own targets (I broke out `browser.js` so that apps could reuse `@haaretz/htz-react-base/browsers` in that case, if they want to use the same list).

Maybe we should clear up what the expectations are for all these builds – if Haaretz apps use `esm`, who are we including `esnext` for (and vice versa)?

packages/htz-components/package.json
exogen: Actually think I got it reversed, `jsnext:main` was an(other?) alias for `module`, and some libraries like `redux` and `react-redux` [still use `jsnext:main`](https://github.com/reactjs/react-redux/blob/fd81f1812c2420aa72805b61f1d06754cb5bfb43/package.json#L7).

In either case, I think we want to clearly distinguish it from that, because people already have a certain meaning for it in their webpack configs. So just plain `esnext` is the way to go.

TxHawks: And what about having different builds for browser and SSR? Just a waste of time?

packages/htz-react-base/babel.js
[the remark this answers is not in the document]
TxHawks: No, you got that right

packages/htz-react-base/babel.js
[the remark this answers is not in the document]
exogen: > Also, will we need different builds for `main` and `browser`?

Including response for this one here too because it's the same answer: in my experience it's not common for component libraries to have a UMD build or a `browser` field. Those are typically used if you make a pre-bundled-for-the-browser file in your library – as in, you're expecting people to drop that file itself in a `<script>` tag (and its dependencies in other `<script>` tags, or provided by a different bundle).

But we don't have a use case for that here. On the server, we just want normal CommonJS modules; wrapping them in UMD wouldn't hurt anything, but is unnecessary. On the client, these files will be found by the app's webpack config, and the point of webpack is that it also expects regular CommonJS or ECMAScript modules. Take `react-redux` and `react-fela` for example: no UMD builds, no `browser` field, and in fact no webpack involved at all, because that's the app's concern, not the library's.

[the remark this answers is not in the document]
TxHawks: I stand corrected. Removing all other comments that reffered to UMD

packages/htz-react-base/babel.js
```diff
@@ -1,30 +1,76 @@
...
+// These are always applied, because they are outside the scope of
+// `babel-preset-env` until such transforms are official.
+const presets = ['react']
+const plugins = ['transform-object-rest-spread', 'transform-class-properties']
+const ignore = []
+
+switch (process.env.BABEL_ENV) {
+  // The `commonjs` environment is used to generate `/dist/lib` and will be
+  // used for traditional Node imports, including during server side rendering
+  // (SSR). It should target the minimum supported Node.js version.
```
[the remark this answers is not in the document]
exogen: That's pretty cool, hadn't seen that before. I've seen some projects use `babel-register` on the server, but it was slower and extremely memory intensive. This looks lighter weight though.

Not sure if you're ever expecting any non-Haaretz consumers of your modules, but I think given that the the CommonJS build is still be useful to such folks (who wouldn't be using `@std/esm`), it'll probably stick around anyway.

packages/htz-react-base/babel.js
```diff
@@ -1,30 +1,76 @@
...
+          // The browser list having any slimming effect on the selected
+          // transforms is mostly aspirational due to the UglifyJS requirement,
+          // see below.
+          browsers: require('./browsers'),
+          // This line is necessary as long as webpack uses UglifyJS for
+          // for minification. Certain syntax transforms are necessary even if
+          // every browser in the list supports it, because UglifyJS needs to
+          // support it too. You may notice that removing this line may not
+          // break the build - most likely, that just means the selected
+          // browsers happen to require the same transforms.
```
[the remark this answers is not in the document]
exogen: Adding an issue to potentially explore that or `uglify-es` in the future.

PR #31, opened 2017-08-30
Apply lint, update Prettier config, add root lint/format scripts

exogen:
* Adds root `lint` and `format` scripts that run those tasks on every package at once.
* Remove the `--no-semi` option from Prettier.
* Fixes `lint` script to exclude `.next` and `dist` directories.
* Fixes some of the outstanding lint issues, but leaves most to sort out soon.


packages/haaretz.co.il/api/articles.js
```diff
@@ -1,11 +1,12 @@
-import 'isomorphic-fetch'
+/* globals fetch */
```
TxHawks: Is this how we want to solve this?

packages/haaretz.co.il/components/Logo/Logo.js
```diff
@@ -1,19 +1,42 @@
...
 
 const style = props => ({
   fill: props.invert ? '#fff' : '#222',
-  color: props.invert ? '#fff' : props.duotone ? '#09A5D9' : '#222'
-})
+  color: props.invert ? '#fff' : props.duotone ? '#09A5D9' : '#222',
+});
 
-/**
+const propTypes = {
```
TxHawks: Since `propTypes` and `defaultProps` describe how to use the component, I rather have them defined at the top like this, rather than at the bottom of the file. 

For class components, I think the best way is defining them as a `static` property.

Agreed?

packages/htz-react-base/eslint.js
```diff
@@ -1,3 +1,87 @@
+/* eslint-disable quotes, comma-dangle */
+/* prettier-ignore */
 module.exports = {
-  extends: 'airbnb'
-}
+  /* Extend airbnb's style-guide enforcement */
+  extends: "airbnb",
+
+  parser: "babel-eslint",
```
TxHawks: Is this a good idea?

It's the only way I could find to support static props. Can change it back to the default parser once support for static props land

packages/haaretz.co.il/layouts/MainLayout.js
```diff
@@ -1,8 +1,23 @@
-import { StyleProvider } from '@haaretz/htz-components'
+import React from 'react';
+import PropTypes from 'prop-types';
+import { StyleProvider, } from '@haaretz/htz-components';
```
TxHawks: I'm not sure why, but `eslint/import` doesn't handle scoped components for some reason.
It throws these two errors:
```sh
3:32  error  Unable to resolve path to module '@haaretz/htz-components'  import/no-unresolved
3:32  error  Missing file extension for "@haaretz/htz-components"        import/extensions   
```

Couldn't find anything about it online.

packages/htz-react-base/eslint.js
exogen: Yep, this is the typical setup.

packages/haaretz.co.il/api/articles.js
exogen: This is fine, if it gets annoying then you could add it to `globals`, but I don't think very many modules will be fetching. For things like `setTimeout` etc. we could use the `shared-node-browser` env.

packages/haaretz.co.il/components/Logo/Logo.js
exogen: I prefer them at the top too. I'm hoping if someone forgets they need to also set `propTypes` at the bottom for SFCs, they'd get an `unused local` error from ESLint?

exogen: Actually, we could just disallow arrow function components – since function statements are hoisted, you could still set `Logo.propTypes = { ... }` at the top and have the function at the bottom.

packages/haaretz.co.il/server.js
```diff
@@ -1,66 +1,67 @@
...
-    })
+    server.get('*', (req, res) => handle(req, res));
 
     server.listen(PORT, err => {
-      if (err) throw err
-      console.log(`> Ready on http://localhost:${PORT}`)
-    })
+      if (err) throw err;
+      // eslint-disable-next-line no-console
+      console.log(`> Ready on http://localhost:${PORT}`);
```
exogen: I would consider just disabling `no-console` altogether, it's one of the least useful rules. Nothing wrong with logging. If you want to strip log messages in the prod bundle, I think there's a minifier option.

packages/haaretz.co.il/api/articles.js
TxHawks: I think it's already in the Airbnb preset, but I'll make sure later.

packages/htz-components/src/components/StyleProvider/renderer.js
```diff
@@ -1,13 +1,14 @@
-import { createRenderer } from 'fela'
-import webPreset from 'fela-preset-web'
+import { createRenderer, } from 'fela';
+import webPreset from 'fela-preset-web';
+import validator from 'fela-plugin-validator';
```
exogen: This `require()` was scoped to that `if` so that this module isn't pulled into the prod bundle, since it's only enabled in dev.

packages/haaretz.co.il/server.js
TxHawks: :+1

packages/htz-components/src/components/StyleProvider/renderer.js
TxHawks: Shouldn't we use `import()`?

packages/htz-react-base/eslint.js
```diff
@@ -1,3 +1,87 @@
+/* eslint-disable quotes, comma-dangle */
+/* prettier-ignore */
```
exogen: Why all the rule differences? I'd be confused why the quotes etc. are different.

packages/htz-react-base/package.json
```diff
@@ -14,12 +14,13 @@
     "htz-scripts": "./bin/htz-scripts.js"
   },
   "scripts": {
-    "format": "./bin/htz-scripts.js format",
+    "format": "./bin/htz-scripts.js format && ./bin/htz-scripts.js lint --fix",
```
exogen: What if we make this the default to cut down on `scripts` complexity, and add a `--no-fix-lint` option to `scripts/format.js`?

packages/htz-react-base/scripts/lint.js
```diff
@@ -1,18 +1,28 @@
-const args = process.argv.slice(2)
+const args = process.argv.slice(2);
 
 // Unfortunately, eslint doesn't just lint the current directory by default -
 // it requires a path to lint. To know if we should add '.' to the argument
 // list, check if any args were passed. If any arg whatsoever was passed, then
 // the user must also specify the path they wish to lint, instead of us adding
 // the current directory by default.
 if (!args.length) {
```
exogen: TODO: Allow `--fix` to be passed without also having to pass a path.

packages/htz-react-base/styleguide.js
```diff
@@ -1,36 +1,40 @@
-const resolveFrom = require('resolve-from')
+/* eslint-disable global-require, import/no-dynamic-require */
```
exogen: I wonder if we could enable this globally in just `htz-react-base`'s `.eslintrc` or `package.json`, since the nature of this package is to do this a lot.

packages/htz-components/src/components/StyleProvider/renderer.js
exogen: I don't think it's possible, `import()` is async isn't it? Whereas we need to export the validator with plugins synchronously.

packages/htz-react-base/package.json
exogen: (By default I mean `format` will also run `lint --fix` by default, so there's just one script call here.)

packages/htz-react-base/eslint.js
TxHawks: Just so it's JSON-like, but you're right, it's stupid

packages/htz-react-base/package.json
TxHawks: 👍

packages/htz-react-base/styleguide.js
TxHawks: We can, the same way I did with the test files - the overrides are a path relative to the config file, so if we do `./**/*.js`, it will only catch files in `htz-react-base`

packages/haaretz.co.il/components/Logo/Logo.js
TxHawks: If someone forgets, ESLint will complain. But it will also complain when trying to use hoisted variables before they are declared (and rightly so, it's bad practice that can easily lead to bugs)

packages/haaretz.co.il/layouts/MainLayout.js
exogen: Hm, might have something to do with Lerna symlinking as well.

packages/htz-react-base/scripts/lint.js
```diff
@@ -1,18 +1,28 @@
...
-require('eslint/bin/eslint')
+// If no --ignore-pattern parameter is specified, use this default.
+// The typical `eslintignore` files and locations will be considered in addition
+// to this parameter.
+if (!args.some(arg => arg.match(/^--ignore-pattern(=|$)/))) {
+  process.argv.push('--ignore-pattern', '.next/*');
+  process.argv.push('--ignore-pattern', 'dist/*');
+  process.argv.push('--ignore-pattern', 'node_modules/*');
+}
+process.argv.push('./**/*.js');
```
exogen: This should already be taken care of above with the `process.argv.push('.')`.

exogen:
I just refactored the `format` script a bit. By default it runs both `format/prettier.js` and `format/eslint-fix.js` (which itself just delegates to the `lint` script). The `eslint-fix` script suppresses the usual lint errors and exit codes since the task is just focused on formatting, unless there's a fatal error (unable to parse).

If `--no-lint-fix` is passed, it just runs Prettier.

packages/haaretz.co.il/api/articles.js
exogen: Another option in this particular case, `import fetch from 'isomorphic-fetch'` so it's not a global.

packages/haaretz.co.il/layouts/MainLayout.js
TxHawks: It does. See [benmosher/eslint-plugin-import#825 (comment)](https://github.com/benmosher/eslint-plugin-import/issues/825#issuecomment-321789702).

Went away when I `rm -rf node_modules && yarn`ed in `packages/haaretz.co.il`.

I'll just add an `import/no-unresolved': { ignore: [ '^#haaretz/.*' ] }` rule. 
`import/extensions` fails for the same reason, and doesn't have an `ignore` option, so I'll completely turn it off .

Not 100% safe, but as good as it gets, I guess.

packages/htz-components/src/components/StyleProvider/renderer.js
TxHawks: And how would that work in the browser? Webpack?

TxHawks:
I think this is good to merge other than the issue of [how to include `fetch`](#pullrequestreview-59634471) 
and the [global require issue](#pullrequestreview-59641721) in `StyleProvider/renderer`.

packages/htz-components/src/components/StyleProvider/renderer.js
exogen: Yup! All imports will be orchestrated by webpack to work in the browser.

TxHawks: :+1:
Let's revert it then

packages/haaretz.co.il/api/articles.js
exogen: I can put it in the server entrypoint, it is common to load them in one place like you suggest.

> so we avoid sending a fetch polyfill to browsers that don't need it

Just so we're clear, doing this will currently avoid sending it to *all* browsers (unless a third-party library pulls it in), even the ones that need it. Would you just take care of covering IE11 in the future and not worry about it now? (e.g. with polyfill.io)

packages/htz-components/src/components/StyleProvider/renderer.js
exogen: I think for the time being we should just leave this as you have it here, and make sure we're not unnecessarily shipping `fela-plugin-validator` in the prod bundle later.

exogen: OK, doing that now

packages/htz-components/src/components/StyleProvider/StyleProvider.js
```diff
@@ -1,16 +1,14 @@
...
-    return (
-      <Provider renderer={styleRenderer}>
-        {this.props.children}
-      </Provider>
-    )
-  }
+export default function StyleProvider() {
+  return (
+    <Provider renderer={styleRenderer}>
+      {this.props.children}
```
exogen: This broke the component since there's no `this.props` anymore, just the function argument, pushing a fix now.

packages/htz-react-base/scripts/lint.js
exogen: .kipped this since `format/eslint-fix` is just part of the `format` command now.

PR #32, opened 2017-08-30
Add dependency sync script

### LINKED ISSUE #7: Handling common cross-dependencies
I've just had an idea, that's probably too naive to work:
What if we include a list up common dependencies, at the `dependencies` field of the root `package.json`, and have a script to install and a script to update these dependencies either as `peerDependencies` and `devDependencies` for consumable packages, or `dependencies` for application packages

This approach is so simple, that I'm sure it has some shortcomings I am not thinking about, because I haven't seen anyone using or suggesting it in the many discussions about this topic.

exogen:
Adds  a `sync` script and a few `syncDependencies` defaults.

Fixes #7.


packages/htz-react-base/package.json
```diff
@@ -35,10 +35,12 @@
...
     "prettier": "^1.5.3",
-    "react": "^15.6.1",
-    "react-dom": "^15.6.1",
     "react-styleguidist": "^6.0.20",
     "resolve-from": "^3.0.0",
     "webpack": "^3.5.5"
+  },
+  "devDependencies": {
+    "react": "^15.6.1",
+    "react-dom": "^15.6.1"
```
exogen: I deliberately had `react` and `react-dom` as real `dependencies` for `htz-react-base` before, so that the Styleguide app running from `htz-react-base` would always use its version. But it makes sense to use the consuming package's version.

I didn't add `peerDependencies` here because `htz-react-base` itself doesn't really require `react`, `react-styleguidist` does, and it should emit whatever `peerDependencies` warnings it needs to the upstream consuming package, so I think it'd be redundant here.

package.json
```diff
@@ -18,7 +19,33 @@
     "url": "https://github.com/Haaretz/htz-frontend/issues"
   },
   "homepage": "https://github.com/Haaretz/htz-frontend#readme",
+  "syncDependencies": {
+    "@haaretz/!(htz-react-base)": {
```
exogen: Everything except `htz-react-base` itself should get `htz-react-base`.

packages/htz-react-base/package.json
exogen: (Also, I don't think htz-react-base doesn't even really need these `devDependencies` since the styleguide will never be run on this package itself, they're here to shut up `react-styleguidist` during dev installs.)

TxHawks:
As a general rule of thumb, I think the tree should be as simple as possible. As long as it does not harm (i.e., creates conflicts, bugs, etc), I rather some packages have `devDependencies` and keep the tree simple and easy to manage. 

Common `devDependencies` will be hoisted by Lerna anyway, so there isn't any significant impact I can think of

TxHawks:
Oh, and we need `prop-types` as a dependency in all packages. 

Do you understand why they recommend to depend on it as a dependency rather than dev+peer in libraries?

exogen:
For things like React, the entire dependency graph needs to see the exact same instance of React for the app to work, or components wouldn't be instances of the same React.Component class. But since `prop-types` is just a handful of validation functions, it doesn't particularly matter how many instances there are. Bundle size is the only concern – but that applies to any module in the dependency graph, `prop-types` doesn't really stand out as more of a concern than everything else.

That said, it's not going to hurt anything if Haaretz modules all go the devDep + peerDep approach. Any third-party component libraries would most likely still pull it in though.

Here's a brief discussion about it (doesn't really add much insight): https://github.com/facebook/prop-types/issues/44

package.json
```diff
@@ -18,7 +19,36 @@
...
+        "react": "^15.6.1",
+        "react-dom": "^15.6.1"
+      }
+    },
+    "@haaretz/htz-components": {
+      "peerDependencies": {
+        "react": "^15.6.1 || ^16.0.0-0"
+      }
+    }
+  },
```
TxHawks: The pattern of doing it by package name rather than by path is problematic, because it basically forces us to manage common package dependencies manually for each package we add to the repo. The path approach was much more efficient.

The current approach get incredibly unwieldy very fast. For example, in if at some point we end up splitting `@haaretz/htz-components` to individual component packages.

scripts/sync.js
```diff
@@ -0,0 +1,46 @@
+/**
+ * Sync dependencies as specified in the `syncDependencies` command of the root
+ * `package.json` file.
+ */
+const path = require('path')
+const spawn = require('cross-spawn')
+
+const rootPackage = require('../package.json')
+const syncDependencies = rootPackage.syncDependencies || {}
```
TxHawks: In order to manage dependencies using paths instead of package names, Use a reducer of  `Object.keys(rootPackage.syncDependencies)` here, which will converts paths to package names

packages/libs/htz-react-base/package.json
```diff
@@ -37,10 +37,15 @@
...
     "react-styleguidist": "^6.0.20",
     "resolve-from": "^3.0.0",
     "webpack": "^3.5.5"
+  },
+  "devDependencies": {
+    "react": "^15.6.1",
+    "react-dom": "^15.6.1"
+  },
+  "peerDependencies": {
+    "react": "^15.6.1 || ^16.0.0-0"
```
exogen: I gave in and just made this package use the same setup as `htz-components`.

exogen:
This was updated to move the packages into `apps`, `components`, and `libs`. The sync script is simpler now and doesn't have to call out to `lerna exec`.

TxHawks:
Lgtm

scripts/sync.js
```diff
@@ -0,0 +1,33 @@
+/* eslint-disable import/no-extraneous-dependencies */
```
exogen: I think this can potentially go in `scripts/.eslintrc`, since it will apply to every file in there. Can do this later.

PR #35, opened 2017-09-01
Fix NODE_ENV and BABEL_ENV in test script

### LINKED ISSUE #34: Test script is failing
Running `yarn run test` in `htz-components` throws with:
```
 FAIL  src/components/Button/Button.test.js
  ● Test suite failed to run

    /home/txh/code/htz-frontend/packages/components/htz-components/src/components/Button/Button.test.js:1
    ({"Object.<anonymous>":function(module,exports,require,__dirname,__filename,global,jest){import React from 'react';
                                                                                             ^^^^^^
    SyntaxError: Unexpected token import

      at ScriptTransformer._transformAndBuildScript (node_modules/jest-runtime/build/ScriptTransformer.js:289:17)
      at handle (node_modules/worker-farm/lib/child/index.js:44:8)
      at process.<anonymous> (node_modules/worker-farm/lib/child/index.js:51:3)
      at emitTwo (events.js:106:13)
```

exogen:
Since the Babel config now requires a `BABEL_ENV` to know what to do, and Jest runs the tests directly in Node (a CommonJS environment), we need to tell the test script to use `BABEL_ENV=commonjs`. 

Fixes #34.

### DIFF
--- a/packages/libs/htz-react-base/scripts/test.js
+++ b/packages/libs/htz-react-base/scripts/test.js
@@ -2,6 +2,9 @@
 const path = require('path');
 const jest = require('jest');
 
+process.env.NODE_ENV = 'test';
+process.env.BABEL_ENV = 'commonjs';
+
 const packageInfo = require(path.join(process.cwd(), 'package.json'));
 
 const argv = process.argv.slice(2);


exogen:
Just a note: after making this change and running the tests again, I was baffled when it kept failing anyway, but started working when I added some `console.log` statements. I believe this is related to Jest caching. If it still fails for you, try running with `yarn run test -- --no-cache`.

exogen:
From the Jest docs:

> Note: a transformer is only ran once per file unless the file has changed. During development of a transformer it can be useful to run Jest with `--no-cache` or to frequently delete Jest's cache.

That means even only the test script or Babel config changes, the cached file won't be updated. 😕

Issue #36, opened 2017-09-01
Discuss Lerna styling approaches for docs


exogen:
For the upcoming documentation, I'm going to include our recommended approach to doing various things with Fela. There are several ways for components to end up with the right `className` using `react-fela`. I'd like to get thoughts on @TxHawks' preferences for the different approaches.

Let's use the current `Logo` implementation as an example. It's currently written in a way that I think betrays my inexperience with Fela, but let's go over it first anyway:

### Direct `renderer` access

This requires setting `contextTypes`, which is a bummer.

```jsx
const style = () => { /* a Fela rule, omitted in remaining examples. */ };

const Logo = (props, { renderer, }) => {
  const className = renderer.renderRule(style, props);
  return (
    <svg className={className} ...snip...>
      <path d="...snip..." />
      <path d="...snip..." />
    </svg>
  );
};

Logo.contextTypes = { renderer: PropTypes.object, };
```

### Easier `renderer` access via a HoC

In order to provide a nicer API when direct `renderer` access is necessary, it's typical to make a HoC so that components can access things from `props` instead. That way people don't have to worry or even know about `contextTypes`. That would look like this (roughly):

```jsx
// This would live in `htz-components`:
function withRenderer(Component) {
  function WithRenderer(props, { renderer }) {
    return <Component renderer={renderer} {...props} />
  }
  WithRenderer.contextTypes = { renderer: PropTypes.object, };
  return WithRenderer;
}

// Use like so:
const Logo = withRenderer((props) => {
  const className = props.renderer.renderRule(style, props);
  return (
    <svg className={className} ...snip...>
      <path d="...snip..." />
      <path d="...snip..." />
    </svg>
  );
});
```

### Using `createComponent`

I didn't want to use the `createComponent` API for the `Logo` class because I didn't like having the intermediate single-use-only `LogoSvg` component. That would look like this:

```jsx
const LogoSvg = createComponent(style, 'svg', ['viewBox', 'width', 'height']);

const Logo = (props) => {
  // Note, don't have to worry about `renderer`.
  return (
    <LogoSvg {...props} ...snip...>
      <path d="...snip..." />
      <path d="...snip..." />
    </LogoSvg>
  )
}
```

### Using `connect`

I prefer this approach to both `createComponent` and `withRenderer`.

```jsx
const Logo = ({ styles }) => {
  return (
    <svg className={styles.container} ...snip...>
      <path d="...snip..." />
      <path d="...snip..." />
    </svg>
[...]

Issue #18, opened 2017-08-28
Server side rendering


TxHawks:
- [ ] Node infasructure
- [ ] Caching strategy
- [ ] Interface with CDN

More?

TxHawks:
@tomerbanin - Please review this to see if anything here is incorrect, inaccurate or incomplete

## Interface with CDN
Currently the CDN keeps a relatively short cache of content elements (lists, teasers, articles, etc.), and hits our servers whenever a request is made after cache is expired to check if content changed and refresh it. This places quite a lot of strain on our servers, and is not ideal in terms of users always getting the latest content.

Instead, we'd like to move to move from this polling model a push model, where when content is created or edited in the CMS, it is pushed to the node server, and then to CDN. We need to see how efficient a mechanism we have to deal with cache invalidation by the node server in the CDN.

@tomerbanin, how fast can we find out and answer this?

exogen:
From the description & discussion, I want to make sure I understand the workflow & needs here:

- The CDN needs both JSON payloads and pre-rendered pages using the same data.
- Since those pre-rendered pages will be output by the server, the responsibility of sending both to the CDN will be the server's.
- Some URLs will be mapped directly to the CDN's copy of those pre-rendered pages, and the node server won't be involved in those (initial) requests.

tomerbanin:
Right now we are basically working with two cache layers :
CDN - Caching data by type and predefined configuration,e.g static files are cached for several days, articles and homepage cached for two minutes.
Varnish - Responsible of fetching data from real front servers and send it back to CDN (by demand), Using also cache settings that are quite similar to those on CDN.

@exogen 
The CDN needs both JSON payloads and pre-rendered pages using the same data.
-- Excactly, CMS server will send JSON payload to CDN for caching and Node server will also send pre-rendered pages for CDN caching.

Some URLs will be mapped directly to the CDN's copy of those pre-rendered pages, and the node server won't be involved in those (initial) requests.
-- Basically as @TxHawks mention before we will prefer setting a long CDN caching for articles (days), as there is no reason not serving an unchanged article from CDN cache. I will check with CDN support for feasibility using re validate cache API (hopefully for a POC next week).

[...]

PR #37, opened 2017-09-01
feat(htz-components):Make consumer responsible for creating the renderer

TxHawks:
As discussed in [#36 (comment)](https://github.com/Haaretz/htz-frontend/issues/36#issuecomment-326676611), this PR is meant to allow the creation of the `styleRenderer` in the app instead of the component lib, so that it can be modified to fit the app's needs.

Specifically, this is meant to help making some of the more gnarling parts of RTL <-> LTR authoring a bit easier.


packages/components/htz-components/src/components/StyleProvider/StyleProvider.js
```diff
@@ -13,10 +22,16 @@ const defaultProps = {
 /*
  * Provider component that makes the Fela renderer available via `context`.
  */
-export default function StyleProvider({ children, }) {
+export default function StyleProvider({
+  children,
+  renderer,
+  ...props
+}) {
```
TxHawks: What I'd really like to do is actually create the styleRenderer here, so that the arguments it is created with are derived from component props, and then pass it to `Provider`.:
```js
const providerRenderer = renderer || createRenderer(isDev, isRtl, selectorPrefix)
<Provider renderer={providerRenderer}>
  ...
<Provider>
```
Unfortunately, I can't really see any way to do this, since we need the `StyleRenderer` before, in [`<Document>`](https://github.com/Haaretz/htz-frontend/pull/37/files#diff-61af6ec6ca789edb986f14c8b7420fd0R4)

packages/apps/haaretz.co.il/components/styleRenderer/styleRenderer.js
```diff
@@ -0,0 +1,5 @@
+import createRenderer from '@haaretz/htz-components';
```
exogen: Should be `{ createRenderer }`?

packages/components/htz-components/src/components/StyleProvider/createRenderer.js
```diff
@@ -0,0 +1,89 @@
...
+    // Auto-prefix properties and values
+    placeholderPrefixer(),
+    prefixer(),
+    // Treat number primitives as a specific css unit
+    unit('rem', {
+      border: 'px',
+      'border-bottom': 'px',
+      'border-left': 'px',
+      'border-right': 'px',
+      'border-top': 'px',
```
exogen: This might support both `border-top` (etc.) and `borderTop`, but the docs (and most CSS-in-JS libraries) use the JS property name, `borderTop`.

packages/components/htz-components/src/components/StyleProvider/StyleProvider.js
```diff
@@ -13,10 +22,16 @@ const defaultProps = {
...
+  /* This allows us to pass down props */
+  const child = Children.only(children);
   return (
-    <Provider renderer={styleRenderer}>
-      {children}
+    <Provider renderer={renderer}>
+      {isValidElement(child) ? cloneElement(child, { ...props, }) : child}
```
exogen: Having a hard time seeing how this would be useful; usually you do this in cases where the component (`StyleProvider` in this case) (1) doesn't own the children, but wants to add extra props (but it's not adding any)
 or (2) does own the children (it doesn't) but wants to support adding extra props to them from the "outside".

That would look like this:

```jsx
  <StyleProvider render={...} bar="baz">
    <Foo /> {/* Foo gets cloned with `bar` prop? Why? */}
  </StyleProvider>
```

Since the rendering component owns the children, they'd just add props there:

```jsx
  <StyleProvider render={...}>
    <Foo bar="baz" />
  </StyleProvider>
```

packages/components/htz-components/src/components/StyleProvider/StyleProvider.js
exogen: Yeah, it really seems like the equivalent of e.g. `createStore()` in Redux, which people do at the top level, before any components even come into the picture. (Of course it's possible to do it within a component, but that would make SSR very difficult.)

packages/components/htz-components/src/components/StyleProvider/createRenderer.js
```diff
@@ -0,0 +1,89 @@
...
+  ];
+  const enhancers = [];
+
+  /* eslint-disable global-require */
+  if (isDev) {
+    plugins.push(require('fela-plugin-validator').default());
+    enhancers.push(
+      require('fela-beautifier').default({ autosemicolon: true, }),
+      require('fela-perf').default(),
+      require('fela-statistics').default()
```
exogen: I may have been wrong about webpack being able to exclude these automatically in production mode, I can't remember if it's smart enough. We can use `IgnorePlugin` to exclude them either way if you feel like moving these back to regular imports.

packages/components/htz-components/src/components/StyleProvider/createRenderer.js
TxHawks: Old habits die a hard...

packages/components/htz-components/src/components/StyleProvider/StyleProvider.js
TxHawks: I think (2) may be useful in testing. Do you see any harm in keeping it?

packages/components/htz-components/src/components/StyleProvider/createRenderer.js
TxHawks: Add reference to this in #18 so that we don't forget

TxHawks:
@exogen - 
I think this is good to merge other than your [comment](#pullrequestreview-60260261) about `StyleProvider` passing props to children.

Like I said, I think it could be useful to be able to pass down props through `StyleProvider`, but I may be wrong. If you think there is actual harm in doing this, I don't mind reverting and should the need actually arise consider adding it later.

packages/components/htz-components/src/components/StyleProvider/StyleProvider.js
exogen: I may be misunderstanding, but the thing is that (2) doesn't really apply because StyleProvider doesn't own any children. No matter what you'd be trying to test, the component rendering the StyleProvider also owns the children and thus can already pass them props directly.

exogen: Definitely won't hurt anything, I just think it should be "as simple as possible until you need it" :)

TxHawks: Couldn't remember where I saw and borrowed this from, but now managed to find it:
https://github.com/cloudflare/cf-ui/pull/149/commits/d939fe521e656dd877112e0cda7dd64e10f090d4

I have no real objection to removing it for now though

packages/components/htz-components/src/document.js
```diff
@@ -1,46 +1,52 @@
...
+ * @param {object} styleRenderer - a react-fela renderer
+ *
+ * @return {class} A Next.js `document` component
  */
-export default class HaaretzDocument extends Document {
-  static getInitialProps({ renderPage, }) {
-    const page = renderPage();
-    const sheetList = renderToSheetList(styleRenderer);
-    styleRenderer.clear();
+const createDocument = styleRenderer =>
```
exogen: Another option here now that I've thought about it: it might be nice to have things like the `styleRenderer` configured in one place – either an `App` instance that contains objects like this, or config values that point to modules to resolve (e.g., one containing the app's preferred `styleRenderer`).

`HaaretzDocument` itself could do something like this, avoiding the need for `createDocument`:

```js
import Config from '@haaretz/app-config';

...
// If it's a module path...
const require(Config.styleRenderer) 

```

`@haaretz/app-config` could do something simple like:

```js
module.exports = require(path.join(process.cwd(), 'app.config.js'));
```

(But we'd need some webpack config to make it OK with that)

exogen:
- [ ] Make sure the styleguides for `htz-components` and `haaretz.co.il` still work and pull in the right style renderer

TxHawks:
> Make sure the styleguides for htz-components and haaretz.co.il still work and pull in the right style renderer

Surprise, surprise, it does break. 
We pass `StyleProvider` as `Wrapper` in `styleguide.config.js`, but it has no renderer since it needs to get in as a prop. We could create a `StyleGuideStyleProvider` component to wrap `StyleProvider` and give it a `renderer` but this seems a bit crooked to me. 

Can you think of a better way around this?

packages/libs/htz-react-base/styleguide.config.js
```diff
@@ -13,6 +13,14 @@ module.exports = {
   title: 'Haaretz Components',
   showCode: true,
   showUsage: true,
+  ignore: [
+    '**/__tests__/**',
+    '**/*.test.js',
+    '**/*.test.jsx',
+    '**/*.spec.js', '**/*.spec.jsx',
+    '**/createRenderer.js',
+    '**/styleRenderer.js',
```
TxHawks: This ignores files that match the `components` glob, but don't actually
export components to avoid annoying warnings during styleguide build.

This is done manually right now on a per-file base. This is okay for
the time being, since we only have two such files, but could become unwieldy
very fast. There must be a better approach for this. Perhaps a file naming scheme?

exogen: A few options I can think of:

1. The real components all have Uppercase names, that could be the convention – nice and simple.
2. There's a `skipComponentsWithoutExample` option where it'd ignore anything without a Markdown file.
3. There's a `logger` option where we could just suppress warnings (or even intercept that particular warning).

PR #39, opened 2017-09-02
feat(haaretz.co.il): Set up an Apollo based app

exogen:
This updates the `haaretz.co.il` app to use Apollo for data fetching. It includes a demonstration of rendering slots based on the JSON page data. The goal is to discuss and iterate on this quickly so we can make sure this is a suitable approach you'll be happy with.

The GraphQL engine runs completely in the browser.

To test it:

```console
$ git checkout wip/apollo-based-store
$ yarn run bootstrap
$ cd packages/apps/haaretz.co.il
$ yarn run dev
```

You may notice that it takes longer than it should to load a second page. That's because in dev mode, none of the page bundles are prebuilt – it's running webpack just-in-time to serve your request. In production mode, they're all prebuilt:

```console
$ yarn run build
$ yarn run start
```


packages/apps/haaretz.co.il/lib/withData.js
```diff
@@ -0,0 +1,212 @@
...
+  refetchSlots.map(name => ({
+    query: ChangedSlotContent,
+    variables: { name, },
+  }))
+);
+
+export default Component => {
+  const componentName = Component.displayName || Component.name || 'Unknown';
+
+  return class WithData extends React.Component {
```
exogen: This whole HoC is based on the official Next-with-Apollo demo, every unofficial demo I've seen uses it as well.

packages/apps/haaretz.co.il/components/Breadcrumbs/Breadcrumbs.js
```diff
@@ -0,0 +1,52 @@
...
+import PropTypes from 'prop-types';
+import Link from 'next/link';
+import { graphql, gql, } from 'react-apollo';
+
+const propTypes = {
+  /**
+   * Information about the current GraphQL query.
+   */
+  /* eslint-disable react/forbid-prop-types */
+  data: PropTypes.object.isRequired,
```
exogen: Apollo controls the shape of this object and I'd like to avoid defining all its fields over and over again. Maybe create a shared PropType for this.

packages/apps/haaretz.co.il/lib/withData.js
```diff
@@ -0,0 +1,224 @@
...
+     */
+    invalidatePage() {
+      this.apolloClient
+        .mutate({
+          mutation: ChangePage,
+          variables: {
+            ...this.props.url.query,
+            pathname: this.props.url.pathname,
+          },
+          refetchQueries: pageRefetchQueries,
```
exogen: A good question to ask is why this is necessary. It's because the `page` field evaluates to the "current page" –
 there's no input argument (e.g. the URL or `contentId`) that components are passing in order to know what data `page` should get. If `page` were parametrized by the URL or `contentId` to fetch, then this wouldn't be necessary and GraphQL would know what to do (keep separate values for each `page` by `contentId` in the store).

If there were a GraphQL server, that's exactly how it would need to work. For the time being, I found it too annoying for each component to know & specify the URL or `contentId`; it made the queries uglier. Something to explore.

packages/apps/haaretz.co.il/lib/withData.js
```diff
@@ -0,0 +1,224 @@
...
+      }
+
+      return {
+        serverState,
+        serverError,
+        ...initialProps,
+      };
+    }
+
+    constructor(props) {
```
exogen: This occurs on every navigation to a separate "page" in Next terms (e.g. from an `article` to a non-`article`).

packages/apps/haaretz.co.il/lib/withData.js
```diff
@@ -0,0 +1,224 @@
...
+      this.apolloClient = createClient({ initialState: serverState, });
+      // If this component is being constructed, it means that a route change is
+      // in progress - unless this is just the initial bootup after being rendered
+      // on the server.
+      if (!serverState.apollo) {
+        this.invalidatePage();
+      }
+    }
+
+    componentWillReceiveProps(nextProps) {
```
exogen: This occurs upon navigation to the same page type (e.g. `article` to `article`).

packages/apps/haaretz.co.il/components/Breadcrumbs/Breadcrumbs.js
```diff
@@ -0,0 +1,55 @@
...
+  /* eslint-enable react/forbid-prop-types */
+};
+
+const Lineage = gql`
+  query Lineage {
+    page {
+      contentId
+      lineage {
```
TxHawks: Would the query need to include on `page > contentId` if the JSON was exploded, or does this has nothing to do with it?

packages/apps/haaretz.co.il/components/Slot/Slot.js
```diff
@@ -0,0 +1,97 @@
...
+      contentId: PropTypes.string.isRequired,
+      contentName: PropTypes.string,
+      inputTemplate: PropTypes.string,
+      properties: PropTypes.object.isRequired,
+    })
+  ).isRequired,
+};
+
+const defaultProps = {
+  content: [],
```
TxHawks: Should required props have defaults?

packages/apps/haaretz.co.il/components/Slot/Slot.js
```diff
@@ -0,0 +1,97 @@
...
+        contentId
+        contentName
+        inputTemplate
+        properties
+      }
+    }
+  }
+`;
+
+export function Slot(props) {
```
TxHawks: Would this need to be a stateful component for us to check is `shouldComponentUpdate` based on wheather its own data changes, or does Apollo take care of it?

Sorry if this is really obvious, still didn't have time to look into Apollo properly

packages/apps/haaretz.co.il/layouts/MainLayout.js
```diff
@@ -1,18 +1,72 @@
...
+        obTitle
+      }
+    }
+  }
+`;
+
+export function MainLayout({ data, children, }) {
+  if (data.loading) {
+    return <div>Loading…</div>;
+  }
```
TxHawks: Wouldn't this always force a full page re-render when browsing between pages? Or will this only happen when changing layouts?

TxHawks:
Components are styled using inline styles instead of Fela just as quick interim prototyping solution, right?
Maybe mention this in a comment, so it's explicit?

exogen:
> Components are styled using inline styles instead of Fela just as quick interim prototyping solution, right? Maybe mention this in a comment, so it's explicit?

Yep! I think given our timeline, I doubt I'll have much time to focus on real styles much, I've skipped using Fela for placeholder/debug rendering. Using it anyway might be good practice with it, though, so I'll know what patterns & pitfalls to document.

TxHawks:
Absolutely no need to actually style thing, especially since three move to react includes a redesign, so it's a complete waste of time, but I think it is a good idea to try and use Fela just to see there aren't any caveats and pitfalls that aren't obvious

packages/apps/haaretz.co.il/components/Slot/Slot.js
exogen: I do this to allow `<Slot />` but disallow something sending `<Slow content={null} />` (only `undefined` will cause it to use the default). But maybe it should just be explicit and always require it.

packages/apps/haaretz.co.il/components/Slot/Slot.js
exogen: As far as I know Apollo doesn't do anything special for your wrapped component. I tend to make everything a `PureComponent` by default, although in most of these cases that won't be good enough and we'll need some deep equality checking, yes.

packages/apps/haaretz.co.il/components/Breadcrumbs/Breadcrumbs.js
exogen: Hmm I don't understand, could you rephrase?

If you're wondering why every query fetches the page's `contentId` even though the component doesn't need it: I don't want to have to specify it, but it's because Apollo identifies objects via an ID in order to normalize, dedupe, and know which are already fetched in the store. For our case I've told it to always identify by `contentId`, so that needs to come back in the query results. I think it would even complain if it were removed here.

TxHawks: That's exactly what I meant

exogen:
I've spent a bit of time trying to address the open questions here and also running them by some other folks at Formidable, now we concretely see what this looks like with your data.

1. How can we ditch the manual cache invalidation step and leave everything up to Apollo (most likely: add a `route` / `URL` / `pathname` / `contentId` query variable to every query).
2. How can we make each component's query less repetitive (needing to specify `page { contentId 
}` in every query).
3. The official solution to both (1) and (2) involves using Fragments, which I've explored for this...it's OK.
4. Will Haaretz devs actually prefer working with this higher-level, more declarative setup to the more low-level pieces and many decisions that come with Redux? Some control is definitely ceded to Apollo.

I think in the end the development experience would be nicer, but my fear is that we won't adequately address the open questions given our timeline and scope. Redux is more of a known-quantity, despite some shortcomings. @TxHawks very interested to hear how you'd like to proceed!

TxHawks:
@tomerbanin , @eliagrady, @avipk, @EranM6, @kedarguy , tawfq (sorry, couldn't find your GitHub username) - please weigh in on this as well.

My two cents:
I'm by no means well versed in Apollo, but for looking at both the documentation and reading the code @exogen committed so far it this PR, I feel it could be a good fit for us and much simpler to work with that straight up Redux. 

I don't feel that, with our limited state management needs,  ceding some control to Apollo will be very limiting.
While I like the Elm architecture behind Redux and how it forces you to think, I feel Apollo's declarative is a better fit for our team.

To the best of my (admittedly very limited) understanding the benefits of using Apollo will grow even further if we end up developing a server-side qraphql endpoint like both @exogen and @eliagrady suggested at some point.

@exogen, please correct me if I'm wrong, but as far as I can tell, adding Redux later, if and when we actually have a concrete need for it should be pretty easy, and even have very little effect on payload size, since Apollo is already using Redux internally.

If you are unfamiliar with Apollo Fragments, please see [documentation here](http://dev.apollodata.com/core/fragments.html) and [a short tutorial here](http://dev.apollodata.com/react/fragments.html)

exogen:
I pushed a refactor of this which I believe is more idiomatic Apollo usage – no more manual invalidation necessary! See the message in 8529598 for more details.

exogen:
The reason I described Fragments as "just OK" is that there's a downside to this approach: the root query component needs to know all the components it will render all the way down the tree in order to include their fragments. Why this is the standard/acceptable approach, I don't know.

packages/apps/haaretz.co.il/components/Slot/Slot.js
```diff
@@ -0,0 +1,80 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import dynamic from 'next/dynamic';
+import { graphql, gql, } from 'react-apollo';
+
+const StandardArticle = dynamic(import('../StandardArticle/StandardArticle'));
```
exogen: There's no reason for this to be a dynamically loaded component, I just wanted to try it out.

TxHawks:
Unless I'm missing the meaning and implications of this, im not at all certain the root query component _can_ know all the components it will render all the way down the tree.

I see two main possible issues:
1) We can only know the type of pages and slots rendered by `MainLayout` after having received and parsed the JSON.
2) Components within most slots are even trickier. Editors can dynamically pick and choose the conetnt that will be displayed, using which component.

packages/apps/haaretz.co.il/components/Slot/Slot.js
TxHawks: But isn't slot very generic and used for _all_ page types? Surely you don't want to load dozens of components you aren't going to use

exogen:
> I'm not at all certain the root query component can know all the components it will render all the way down the tree

Right, that's why I'm confused that this is the official solution. This doesn't just apply in your case, but I think in general. It doesn't seem clean. I'm actually not too worried about it in your case though, because once we get to `slots`, no components below that are going to need to add fragments to the root query (rather, they'd just make a new query).

> We can only know the type of pages and slots rendered by MainLayout after having received and parsed the JSON.

* Type of page: Are you thinking that possibly we'll have to mostly do without Next's "page type" routing and instead of a globally-used Page type that then delegates to `ArticlePage`, etc. based on the query response? Is `pageType` not always guessable by URL? 
* Slots: I think for slots that's not a problem. I can just change to schema to fetch all slots for the page instead of specifying which to request ahead of time. I figured in the future (if there were a GraphQL server) this would be useful to load the above-the-fold slots initially, then request the rest.

> Components within most slots are even trickier. Editors can dynamically pick and choose the conetnt that will be displayed, using which component.

Not too worried about that, same as the first response – the content elements within slots won't need to add fragments to the root query, but would have their own queries. Hope that all makes sense!

packages/apps/haaretz.co.il/components/Slot/Slot.js
exogen: Depends on the component – each `import()` makes a new webpack chunk, meaning a separate file to request. For most components, it's going to be way snappier and efficient to just include them in the same initial bundle rather then fire off dozens of extra requests, one for each component (extra requests often perform worse than a slightly bigger initial bundle). More likely, you'd want chunks to have related groups of components that are likely to be used together.

TxHawks:
> Type of page: Are you thinking that possibly we'll have to mostly do without Next's "page type" routing and instead have a globally-used Page type that then delegates to `ArticlePage`, etc. based on the query response? Is `pageType` (in your JSON schema) not always guessable by URL?

Unfortunetely, `pageType` is almost never guessable by the url, except for very few exceptions. We will need to configure some routes, e.g., for section pages, which don't expose their `comtentId`, but for most all pages, the path is irrelevant, and the only thing that matters is the `contentId`. The CMS's internal dispatcher will look at what the `contentId` is mapped to, we can't know it beforehand

exogen:
OK, doesn't sound like too much of a problem to me. The routing is the weakest aspect of Next anyway, so we're not losing out on much by needing to defer to the JSON endpoint. I'm going to update this PR to see how that works out.

exogen:
That does require answering a question I've had though: how do we know the `contentId` of some pages like `/` that don't have a `contentId` in their URL? Is it hardcoded/configured in the app, or does the JSON endpoint also have the ability to return data for particular paths like `/`, with no input `contentId`?

TxHawks:
I guess `/` is very simple and should be hard coded into the app, but I don't think we'll want to do that for other pages, e.g., section pages, that don't expose their `contentId`. The endpoint should know how to handle that

exogen:
* Updated to demo Fela styles in 5cd7a2e.
* Removed predefined slots and separate `/article` route in 05d1b2a. The `pageType` is returned in the query, but doesn't currently delegate out to a different component based on it because I don't know what any necessary differences are between page types at the moment.

eliagrady:
@exogen routing in our case starts by asking for the requested URL, whether externally from other links - or by internal links. 
For each such URL, there is a special mode that gives us the page's JSON, which in turn has internal knowledge regarding the page type (from the CMS).
I'll give you an example:
Say a request for page:
http://elia.haaretz.co.il/news/world/asia/.premium-1.5527
was requested.
We ask the backend for ${HOST}/papi/${PATH} to get the JSON for ${HOST}/${PATH}
so in this case, we query:
http://elia.haaretz.co.il/papi/news/world/asia/.premium-1.5527
Each of these requests is for a complete 'page' representation (slots, page type etc.)
Partial data, such as the 'Comments' component, will have an API endpoint for it's own: say a 'Comments' component has a contentId of 7.4335652, then the component will fetch from http://www.haaretz.co.il/json/cmlink/7.4335652 a custom JSON for that element.

TxHawks:
Just to clarify, requesting 
https://www.haaretz.co.il/papi/news/politics/.premium-1.4424682
https://www.haaretz.co.il/papi/news/politics/1.4424682
https://www.haaretz.co.il/papi/.premium-1.4424682
https://www.haaretz.co.il/papi/1.4424682

All will all yield the exact same JSON output

PR #40, opened 2017-09-03
feat(htz-frontend): Add a clean script, ensure 'yarn clean' does no harm

exogen:
* `yarn run clean` will now run `lerna clean --yes` and remove distribution files.
* `yarn run bootstrap` will clean and install, `yarn run update` will just install.
* `yarn clean` does nothing due to the `.yarnclean` file, which specifies file patterns to remove from `node_modules`. (The default settings often break packages that include non-JS assets.)
* Adds `htz-scripts clean`.


package.json
```diff
@@ -6,6 +6,8 @@
   "main": "index.js",
   "scripts": {
     "bootstrap": "lerna bootstrap",
+    "bootstrap:clean": "yarn run clean && yarn run bootstrap",
```
TxHawks: Let's rename this to `restart`? 
Another option is to have this as the default for `bootstrap`, and rename the current `bootstrap` to `update` or something like that

package.json
```diff
@@ -6,6 +6,8 @@
...
+    "clean": "lerna clean --yes",
```
TxHawks: This should probably also `rimraf` everything in all `dist` dirs inside packages and remove the Jest cache

exogen: Sounds good to me. I think I'll add a `clean` command to `htz-scripts` so it can delegate to the individual packages, in case one ends up with custom cleaning needs.

e.g.

`lerna clean --yes && lerna exec -- yarn run clean`

package.json
exogen: I like `bootstrap` and `update`, since "bootstrap" implies from the ground up. 👍

package.json
exogen: Hm wait, that doesn't work. If `lerna clean` was just run, none of the packages have their dependencies, so no `htz-scripts`... hmMmM

exogen: Which of these would you prefer?

* Make `clean:node_modules` and `clean:dist` two separate scripts in the root (`clean:dist` only being able to run any time after `bootstrap`)
* Every package depends on `rimraf`, so `htz-scripts` isn't necessary to `clean`.
* Root package just decides what to remove in every sub-package

TxHawks: I think 3 is good enough. If we ever run into needing something more complex we can refactor then

exogen:
Looks like `lerna bootstrap` isn't aware of the `prepare` lifecycle script, which presents another problem: when you link something like `htz-components` in `haaretz.co.il`, you're counting on `htz-component`'s `dist` files being there, because that's what consuming modules like `haaretz.co.il` actually import. I'll need to switch that to `prepublish` even though `prepare` is a better option AFAIK.

packages/components/htz-components/package.json
```diff
@@ -14,9 +14,10 @@
   "esnext": "dist/esnext/index.js",
   "scripts": {
     "build": "htz-scripts build",
+    "clean": "htz-scripts clean",
     "format": "htz-scripts format",
     "lint": "htz-scripts lint",
-    "prepare": "npm run build",
+    "prepublish": "npm run build",
```
TxHawks: No need to do this if we update to [Lerna 2.1.2](https://github.com/lerna/lerna/releases/v2.1.2)

TxHawks:
Updating Lerna to  [2.1.2](https://github.com/lerna/lerna/releases/v2.1.2) should fix the `prepare` issue (was addressed in Lerna/lerna#979).

We should probably add a note in the root README noting that node 8 is required (Lerna's using the `prepare` script depends on having npm 5 installed).

package.json
```diff
@@ -5,10 +5,12 @@
   "description": "",
   "main": "index.js",
   "scripts": {
-    "bootstrap": "lerna bootstrap",
+    "bootstrap": "yarn run clean && yarn run update",
+    "clean": "lerna clean --yes && rimraf \"packages/*/*/{.next,dist}\"",
```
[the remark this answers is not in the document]
exogen: This is definitely not possible from the root since the `jest` module chooses where it is (a random directory like `/var/folders/41/bzdlyd8j6cbfyj0q1by7jkkh0000gn/T/jest_dx`), unless we switch the order to run `lerna clean --yes` last and hope `node_modules` is installed beforehand.

But I did add it to that `htz-scripts clean` command.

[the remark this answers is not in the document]
exogen: Or I suppose we could also install `jest` in the root project and hope that its cache directory is a global & shared among each package (although each package could also override it).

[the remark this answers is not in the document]
exogen: Maybe our default Jest config could choose a relative per-project directory for it like `./.jest`? Then we could just add it to the `rimraf` above.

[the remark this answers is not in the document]
TxHawks: Jest config has a [`cacheDirectory` option](https://facebook.github.io/jest/docs/en/configuration.html#cachedirectory-string), so we can easily set it to a relative local dir, which would make it predictable and save us the need to worry about installing Jest in the root

TxHawks:
:+1:

PR #41, opened 2017-09-03
feat(haaretz.co.il) Setup basic production-minded server settings

exogen:
* Read environment variables from a `.env` file. Copy the `.env.sample` file to get yourself the defaults. This makes it easy to manage multiple environment variables during local development, or if it's cumbersome to set them during deployment.
* Add [compression](https://github.com/expressjs/compression) and [security](https://helmetjs.github.io/) middleware.


exogen:
I think eventually much of the server logic will be in a shared module (either `htz-react-base` or something new like `htz-react-server`). Same with a lot of the deps currently in `haaretz.co.il`. But I think we should see how things shake out in the first app before trying to guess how to factor it.

TxHawks:
👍 for eventually having server logic encapsulated I. `htz-react-server`. Just opened #42 for this.
LGTM

exogen:
👍  Thanks for tracking it!

PR #45, opened 2017-09-08
docs(all): Add a System Overview document

exogen:
I think I'll be adding more sections to this document, but Request Lifecycle is probably the most useful.

(I recommend clicking the little 📄 icon to see the rendered Markdown in this PR.)


docs/SystemOverview.md
```diff
@@ -0,0 +1,154 @@
...
+  * The Page API client is simply a [`DataLoader`][DataLoader] instance. This
+    utility makes it easy to batch in-flight requests and cache the results.
+  * :link: See the page query in [layouts/MainLayout.js][].
+  * :link: See the `page` resolver in [lib/schema.js][].
+  * :link: See the `DataLoader` instance in [lib/createClient.js][].
+* Apollo saves the results of all the GraphQL queries in its store. This data is
+  then retrieved from the store and returned as one of the props in
+  `getInitialProps`.
+* Once `getInitialProps` is complete, Next.js renders the page component with
+  the returned props using a ReactDOM method like [renderToString][].
```
TxHawks: We should probably add a TODO here (and open an issue) reminding us to consider replacing `renderToString` with `renderToStream` once React 16 is out of beta.
Where in the code does this live?

exogen: We don't do it directly, Next takes care of it: https://github.com/zeit/next.js/blob/75a43f5a7158905b047cfb9a4516dac90a6ffea3/server/render.js#L76-L85

So in the future you'd get streaming support by just upgrading Next (I'm sure there will be plenty of demand for it). You might wait a bit longer to get it, but you can be sure they've thought things through – streaming is a lot tricker than just replacing `renderToString` due to errors (to start streaming to the client you need to send the HTTP response code first, but not all components have tried rendering yet, maybe it should actually be a 404 or 500?). It might be the case that streaming is just used to break up work on the server (for better concurrency – `renderToString` is synchronous and people often focus on trying to optimize it heavily) instead of actually responding to the request faster.

exogen: Added a note!

PR #47, opened 2017-09-09
refactor(haaretz.co.il): Ditch contentId parsing, fetch pages by path

exogen:
Now that I know the Page API will retrieve pages based on their full path, it doesn't seem like we need to parse the `contentId` out of the URL. Here I've replaced the `section`, `tier`, `contentId` params with just `path`. This makes a bunch of code simpler. The page's path can be the query input to the Page API.

If you ever want URL parsing again, you can always resurrect this code later.


TxHawks:
We will need to do _some_ parsing of the path, but I guess we can take care of it later, when it's actually relevant.

To the best of my understanding, we will need to check if the path includes a `CMLink` string, and if not, under `papi` after the first `/`.

I say, let's open an issue noting exactly where this should later be implemented and merge it.
@eliagrady

exogen:
@eliagrady @tomerbanin I think I don't fully understand the workflow described and need it spelled out a little more, but maybe I don't have to since my time is almost up. I opened issue #60 and will merge this for now!

PR #49, opened 2017-09-09
chore(all): Use Yarn Workspaces with Lerna

TxHawks:
[Yarn Workspaces](https://yarnpkg.com/blog/2017/08/02/introducing-workspaces/)
makes hoisting and common dependency management easier and more
streamlined.

Since the `workspaces` feature requires Yarn version >= 1.0.0, this PR also adds an `engines/yarn` field to the root `package.json`


exogen:
Looks like there's one thing left to make things work: currently, trying to `yarn add` a dependency in any of the packages fails with an error: 

> error An unexpected error occurred: "ENOENT: no such file or directory, open '/Users/brianbeck/Projects/Haaretz/htz-frontend/packages/components/htz-components/.yarnclean'".

Copying or symlinking the root `.yarnclean` into each package makes it work.

It looks like either way, `yarn remove` doesn't work in packages anymore. 

> error No lockfile in this directory. Run `yarn install` to generate one.

I guess you'd have to edit `package.json` manually and rerun `yarn` to remove a dependency.

exogen:
Actually, it looks like they realized what a bad idea `yarn clean` was and renamed it to `yarn autoclean`. Since I doubt people will run that by mistake, we can probably remove `.yarnclean` altogether.

PR #51, opened 2017-09-09
feat(htz-frontend): Add a way to lint and format the root package

### LINKED ISSUE #33: Add a way to format and lint scripts dir in the root package
Could be as simple as reaching into `htz-react-base`, like:

```$ node packages/libs/htz-react-base/bin/htz-scripts.js format script```
```$ node packages/libs/htz-react-base/bin/htz-scripts.js lint scripts```

But at a minimum, `format` needs to be updated to allow linting individual directories.

### PR DESCRIPTION
**Pull request by exogen**
I made this as simple as just adding `packages` to the ignore list in the `format` and `lint` options. Non-root packages should never have their own `packages` dir.

Fixes #33.

### CODE REVIEW

File: packages/libs/htz-react-base/scripts/format/index.js [hunk 1]
```diff
@@ -10,8 +10,8 @@ function getFileList() {
   return globby.sync(
     [
       '**/*.{js,jsx}',
-      '!**/{.jest,.next,dist,node_modules}/**',
-      '!./{.jest,.next,dist,node_modules}/**',
+      '!**/{.jest,.next,dist,node_modules,packages}/**',
```
TxHawks: Maybe we can remove `packages` from this line, and then we only ignore the top-level `packages` , which I think is what we acctually want in order to be able to format the files in the root dir.

exogen: Hmm, it seems that the las tline `'!./{.jest,.next,dist,node_modules}/**'` is totally redundant here as `**/` already includes `./`? Can't remember why we have both...

But anyway – unfortunately that won't make it only ignore the root `packages`, because the packages are formatted with `lerna exec --no-bail -- yarn run format`, so they're using this same glob with CWD set to their own package dir.

TxHawks:
Let's merge then

exogen:
Possibly something to revisit later, but I agree we should just go with this now for simplicity.

PR #54, opened 2017-09-10
fix(haaretz.co.il): Styleguide build with Yarn Workspaces

### LINKED ISSUE #52: App styleguide broken by move to Yarn Workspaces
@exogen reported in Slack:
> The app’s styleguide is busted cuz it tries to find components in `node_modules/@haaretz/htz-components`, which is normally a safe assumption, but hoisting breaks this. We need to account for this by doing some path manipulation with `require.resolve('@haaretz/htz-components')` in `styleguide.config.js`, but even that was not working at first because `bootstrap` doesn’t perform lifecycle scripts anymore, so `dist` doesn’t exist after `bootstrap` (can work around it by doing `require.resolve('@haaretz/htz-components/package.json')`, but it’s still annoying to have to go around and build everything after bootstrapping)

### LINKED ISSUE #53: App build broken by move to Yarn Workspaces
@exogen in Slack:
> `build` in the app is broken and spits out a ton of errors (it’s just running a vanilla `next build`, we don’t do anything fancy), I assume because Next can’t resolve things properly when they’re hoisted (edited)

TxHawks:
* Correctly resolve `@haaretz/htz-components` in the app styleguide 

closes #52
closes #53

### DIFF
--- a/package.json
+++ b/package.json
@@ -5,7 +5,7 @@
   "description": "",
   "main": "index.js",
   "scripts": {
-    "bootstrap": "yarn run clean && yarn run update",
+    "bootstrap": "yarn run clean && yarn run update && lerna exec --no-bail -- yarn run build",
     "clean": "lerna clean --yes && rimraf \"packages/*/*/{.jest,.next,dist}\"",
     "docs": "doctoc --title \"**Table of Contents**\" .",
     "format": "lerna exec --no-bail -- yarn run format",
--- a/package.json
+++ b/package.json
@@ -5,7 +5,7 @@
   "description": "",
   "main": "index.js",
   "scripts": {
-    "bootstrap": "yarn run clean && yarn run update && lerna exec --no-bail -- yarn run build",
+    "bootstrap": "yarn run clean && yarn run update && lerna run prepare",
     "clean": "lerna clean --yes && rimraf \"packages/*/*/{.jest,.next,dist}\"",
     "docs": "doctoc --title \"**Table of Contents**\" .",
     "format": "lerna exec --no-bail -- yarn run format",
--- a/packages/apps/haaretz.co.il/styleguide.config.js
+++ b/packages/apps/haaretz.co.il/styleguide.config.js
@@ -1,5 +1,5 @@
 const { configure, } = require('@haaretz/htz-react-base/styleguide');
-const htzComponentsPath = require('path').dirname(require.resolve('@haaretz/htz-components'));
+const htzComponentsPath = require('path').dirname(require.resolve('@haaretz/htz-components/package.json'));
 
 // When passed an object, `configure` will do a (shallow) extend of the default
 // config. If you need to extend a particular value (e.g. `styleguideComponents`),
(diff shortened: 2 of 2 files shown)


packages/apps/haaretz.co.il/styleguide.config.js
```diff
@@ -1,4 +1,5 @@
 const { configure, } = require('@haaretz/htz-react-base/styleguide');
+const htzComponentsPath = require('path').dirname(require.resolve('@haaretz/htz-components'));
```
TxHawks: This is probably better anyway, as its less brittle than including an assumption about the path as a hardcoded string.

packages/apps/haaretz.co.il/styleguide.config.js
```diff
@@ -16,7 +17,7 @@ module.exports = configure({
     {
       name: 'Shared Components',
       components:
-        'node_modules/@haaretz/htz-components/src/components/{,src}/**/[A-Z]*.{js,jsx}',
+        `${htzComponentsPath}/../esnext/{,src}/**/[A-Z]*.{js,jsx}`,
```
TxHawks: Problem with getting components from `dist` is that the `md` files are not copied, so there are no exaples of the shared components in the styleguid.

Two possible options:
1. Copy `md` files to `dist` dir
2. Link to `src` dir instead.

I think I like (1) better. @exogen, what do you think?

packages/apps/haaretz.co.il/styleguide.config.js
exogen: I would resolve `@haaretz/htz-components/package.json`, because it will always point you to the package's root directory instead of wherever the `main` field goes – that way you don't need to make assumptions about where to go from the `main` file.

packages/apps/haaretz.co.il/styleguide.config.js
exogen: You pretty much have to link to `src`, as that's what `react-docgen` is expecting to parse anyway.

package.json
```diff
@@ -5,7 +5,7 @@
   "description": "",
   "main": "index.js",
   "scripts": {
-    "bootstrap": "yarn run clean && yarn run update",
+    "bootstrap": "yarn run clean && yarn run update && lerna exec --no-bail -- yarn run build",
```
exogen: `lerna exec --no-bail -- yarn run build` maybe put this in a `build` script here and make this `&& yarn run build`?

exogen:
We'll have to remember to remove the `&& build` part when Yarn fixes lifecycle scripts (https://github.com/yarnpkg/yarn/issues/3911), because (1) that would make it start running `build` twice, which is slow, and (2) this is just making the assumption that `build` is the right thing to run post-install – different packages may end up doing different things in `prepare`, `prepublish`, etc.

TxHawks:
Maybe we should use `yarn lerna run prepublish` instead for the time being?

exogen:
That sounds like an improvement, even though there are still several different lifecycle scripts that would ideally be run. I'd make it `lerna run prepare` and our development docs should tell package authors to use `prepare`.

TxHawks:
:+1:
Are we currently using prepare in any of the packages?

exogen:
Yup! It's how `htz-components` builds.

exogen:
And I should also note that it's deliberately *not* in `haaretz.co.il` – apps don't really need to build post-install, because they're not dependencies of anything, they're supposed to be built upon deployment (or when you're working on them).

TxHawks:
I'll fix the script in a minute and will remove it when yarnpkg/yarn#3911 is resolved.
Can you add comment about `prepare` to the docs?

PR #59, opened 2017-09-11
feat(all): Add global app configuration and docs

exogen:
* Use the node-config library to load environment-specific configuration files.
* Add a `config` shim for the browser.
* Update Button component to use a demo config value.
* Add a snapshot helper for testing Fela styled components (this was part of using config in the Button demo).
* Fix the default Styleguidist `contextDependencies` now that hoisting is in place.
* Improve error handling in the `format:eslint-fix` script.

Note: GitHub breaks some of the relative file links in the "Rich Diff" mode, use "View" to view the file in this branch if you want the links to work.


packages/libs/htz-react-base/styleguide.config.js
```diff
@@ -84,5 +85,16 @@ module.exports = {
...
     },
+    plugins: [
+      new webpack.DefinePlugin({
+        // The only module that should be accessing this directly is
+        // `./webpack/browserConfig` - this provides an easy way to get this
+        // value into the shimmed `config` module.
+        // NOTE: webpack already has support for stringifying an object value
+        // provided here, but we want to make sure it serializes it the exact
+        // same way that a real app will, using `serialize-javascript`.
+        'window.__HTZ_CONFIG__': serialize(config),
```
exogen: Realistically, we could remove the new `Document` method and just always do it via webpack (using `DefinePlugin` or one of several other reasonable ways). I like how `Document` gives us (1) more control and visibility of exactly where the config will appear, and (2) an easily inspectable script when you View Source, instead of hiding it in the bundle.

packages/libs/htz-react-base/styleguide.config.js
```diff
@@ -62,18 +65,16 @@ module.exports = {
   contextDependencies: [
     path.join(process.cwd(), 'components'),
     path.join(process.cwd(), 'src', 'components'),
-    path.join(
-      process.cwd(),
-      'node_modules',
-      '@haaretz',
-      'htz-components',
-      'src'
-    ),
```
exogen: Doesn't work for hoisted `htz-components` deps, so I just made it the responsibility of the apps that include it.

packages/components/htz-components/src/document.js
```diff
@@ -1,23 +1,31 @@
...
  * `pages/_document.js` file of a Next.js app.
  *
  * @param {object} styleRenderer - a react-fela renderer
+ * @param {object} appConfig - An app config object, defaults to the one the
+ *                             `config` module loads.
  *
  * @return {class} A Next.js `document` component
  */
-const createDocument = styleRenderer =>
+const createDocument = (styleRenderer, appConfig = config) =>
```
exogen: I'm not sure this is necessary, I just thought it might come in handy.

packages/libs/htz-react-base/test/createSnapshotFn.js
```diff
@@ -0,0 +1,14 @@
...
+
+function createSnapshotFn(styleProvider) {
+  return component => ({
+    // Output an object with both the component snapshot and raw styles.
+    component: renderer
+      .create(React.cloneElement(styleProvider, {}, component))
+      .toJSON(),
+    styles: styleProvider.props.renderer.renderToString(),
+  });
+}
```
exogen: This is from the Fela + Jest docs, but a little fancier.

packages/components/htz-components/src/components/Button/Button.test.js
```diff
@@ -1,8 +1,14 @@
...
+import createRenderer from '../StyleProvider/createRenderer';
+import StyleProvider from '../StyleProvider/StyleProvider';
+
+const felaSnapshot = createSnapshotFn(
+  <StyleProvider renderer={createRenderer()} />
+);
 
 it('renders correctly', () => {
-  const tree = renderer.create(<Button>Click here</Button>).toJSON();
+  const tree = felaSnapshot(<Button>Click here</Button>);
```
exogen: This is from the Fela + Jest docs.

package.json
```diff
@@ -32,6 +32,14 @@
     "yarn": "^1.0.0"
   },
   "syncDependencies": {
```
exogen: This is getting really difficult to understand at a glance. Might be better flipped on its head (dependency names are the top-level keys, package globs are the leaf values)? Doubt I'll have time to get to it though.

packages/libs/htz-react-base/styleguide.config.js
TxHawks: I prefer it explicit like that too. The less magic, the better

packages/libs/htz-react-base/eslint.js
```diff
@@ -58,6 +58,8 @@ module.exports = {
       'LabeledStatement',
       'WithStatement',
     ],
+    /* Allow __HTZ_CONFIG__ name */
+    'no-underscore-dangle': [ 'error', { allow: [ '__HTZ_CONFIG__', ], }, ],
```
exogen: I just named it this for symmetry with `__NEXT_DATA__`. Maybe just plain `HTZ_CONFIG` is nicer? Or `HTZ_DATA` in case it needs to contain more than the config object in the future.

packages/components/htz-components/src/document.js
TxHawks: No harm in it, and can be helpful

packages/libs/htz-react-base/eslint.js
TxHawks: I like the dangling underscore to denote the is a meta global. Using `DATA` instead of `CONFIG` is a good idea

exogen: Done! It now looks like `__HTZ_DATA__ = { config: { ... } }`.

Also added a FAQ item.

Issue #58, opened 2017-09-11
Documentation additions and expansion


TxHawks:
- [ ] Mention how to extend Babel, Webpack, Jest ans ESLint configs for apps in the [relevant section](../blob/master/docs/FAQ.md#how-do-i-see-the-configuration-for-babel-webpack-jest-eslint-etc) of the FAQ, e.g., in order to include and configure `babel-plugin-lodash`
- [ ] The [packages section](../blob/master/README.md#packages) of the README should point out how to publish existing packages (all or specific ones), and how to create a new package (including updating the `syncDependencies` property

PR #61, opened 2017-09-12
docs(all): Document how to extend configs

exogen:
One half of #58.

This documents how to extend our configuration files in a specific package, and improves some of the extension capabilities already in place.


docs/FAQ.md
```diff
@@ -83,6 +84,61 @@ scripts].
...
+  config].
+* Note that not all documented config location options are supported by our
+  `lint` script – it will not check for a field in `package.json`, nor several
+  of the alternative file formats.
+* You must manually extend the default config or its rules won’t be included:
+  ```js
+  module.exports = {
+    extends: require.resolve('@haaretz/htz-react-base/eslint')
+  };
+  ```
```
exogen: This is a consequence of the precedence ESLint uses. `--config` is higher precedence than a `.eslintrc` file, and we need to pass `--config` so it knows where the default is. One solution might be to move the default from `htz-react-base/eslint.js` to `.eslintrc.js` in the root, but that might require moving some dependencies there too.

TxHawks: Or, we can always have an `.eslintrc` that only extends the base.

This requires the least mucking around, can easily be replicated by a build script and is easily extendible

exogen: Making sure I understand: by that do you mean have a `.eslintrc` at the repo root, that extends `packages/libs/htz-react-base/eslint`, where all the config options would still live?

TxHawks: No, at the root of each individual package, that extends `@haaretz/htz-react-base/eslint`. This way it's always obvious how to extend the configuration

exogen: Gotcha. It's unfortunate that just plain `extends: '@haaretz/htz-react-base/eslint'` didn't work when I tried it. It tries to add the `eslint-config-` prefix to the module name. Other than that small bit of `require.resolve()` ugliness, sounds like a fine option to me!

TxHawks: :+1:
Let's do it and merge

exogen: Done, although I hadn't thought about the consequences of needing to lint the root package – it now has a `.eslintrc.js` file too so ESLint knows how to lint its scripts. Due to the parent directory traversal, individual packages don't actually need to specify their own `.eslintrc.js` files like you wanted, but I added one in each package anyway.

Thinking about what the best potential combination of config options to use is (`.eslintignore`, `--ignore-pattern`, `overrides`, `root: true`) to get around this. Ideally each individual package would not need to use `root: true` as it's more boilerplate to potentially forget. Or maybe remove the root `.eslintrc` and specify the path manually in `lint:self` (though that's harder to do for `format:self`...)

exogen: Maybe the individual packages can just have `module.exports = {};` in their `.eslintrc.js` files? That way they it's obvious where to put custom config, but they don't really need to put anything for (the root's config points to `htz-react-base` for them).

TxHawks: Does it traverse up even if you do have a local, non-empty one?

exogen: Yeah, the idea is that subdirectories can define overrides for themselves, and it doesn't see the package dirs any different in this regard (unless they have `root: true`).

TxHawks: 👍  then

exogen: Which approach? sorry, not 100% awake yet today

TxHawks: Actually, either one. I say, just merge whatever you already have in place and take less of your time. I don't think it really matters since I plan on having a script that initializes new packages, so there is no issue of forgetting to `extend` or need to do yet another thing manually.

exogen: Gotcha, merging as-is.

PR #62, opened 2017-09-12
fix(htz-react-base): Fix production build by using ES5 syntax in config shim

exogen:
I forgot that nothing in `htz-react-base` gets transpiled currently, and the webpack config doesn't transpile anything under `node_modules` (as is typical), so `let`, `const`, and backticks in the config shim were giving UglifyJS troubles.

Simplest solution for now it to use ES5 syntax.


TxHawks:
Maybe just come figure Webpack to consume and transpilie `esnext` instead of `main` for packages where it is available?

PR #63, opened 2017-09-14
feat(haaretz.co.il): Add a basic Service Worker setup

exogen:
* Add both a development service worker (no precaching, but exists to take control from any other running service worker) and production service worker (precaches JS bundles).
* Use Workbox webpack plugin to generate a `sw.js` file from the different service worker templates.
* Add basic documentation.
* Add logging middleware to the Application Server, this can help when debugging to see which requests are making it to the server.


packages/apps/haaretz.co.il/next.config.js
```diff
@@ -17,6 +18,104 @@ module.exports = {
...
+              const buildId = match[1];
+              const filePath = match[2];
+              const page = match[3];
+              let { url, } = entry;
+              if (filePath === '/app.js') {
+                url = `/_next/${entry.revision}/app.js`;
+              }
+              else if (page === '/_error.js') {
+                url = `/_next/${buildId}/page/_error/index.js`;
+              }
```
exogen: For `app.js` and `_error.js` above, it might actually be redundant to include them in the service worker precache, but I haven't thought it all the way through (service workers and their lifecycle are a little mind bending).

When you hit the page, `app.js` and `_error.js` are some of the first things fetched by Next.js. Then the Service Worker appears to fetch them again, to precache them. I can't think of a scenario where `app.js` and `_error.js` wouldn't have already been fetched...so is this redundant? I'm not sure, because service workers also enable offline support, so maybe it's good for it to also fetch & cache them?

packages/apps/haaretz.co.il/lib/serviceWorker/registerServiceWorker.js
```diff
@@ -0,0 +1,33 @@
...
+        .catch(err => {
+          console.error(
+            '[Application] Service Worker registration failed.',
+            NO
+          );
+          console.error(err);
+        });
+    }
+  }
+}
```
exogen: This could attempt to do smarter things, see the `create-react-app` implementation for reference: https://github.com/facebookincubator/create-react-app/blob/master/packages/react-scripts/template/src/registerServiceWorker.js

But I hate to copy too much code from there that I don't fully understand.

PR #64, opened 2017-09-14
chore: Encapsulate server logic in htz-react-server

### LINKED ISSUE #42: Encapsulate server logic in ”htz-react-server"
Move shared server logic and dependencies from `apps/haaretz.co.il` to `htz-react-server` once it's mature enough

exogen:
This currently offers no extensibility for apps to add their own routing, middleware, or anything like that, but that could be a future enhancement. Maybe it could be entirely config based? At least for routing, `htz-react-server` could probably `import config from 'config'` and add a few `server.get()` calls with routes defined there.

Fixes #42.


packages/libs/htz-react-server/src/index.js
```diff
@@ -0,0 +1,41 @@
+import 'dotenv/config';
+import express from 'express';
+import compression from 'compression';
+import helmet from 'helmet';
+import next from 'next';
+// Even though this isn't used directly in this package, this file acts as the
+// server entry point, so it should make any environment adjustments (like
+// adding this `fetch` global) here. That way they'll be available to any
+// modules that Next.js imports while routing and rendering pages.
+import 'isomorphic-fetch';
```
TxHawks: This isn't loaded in the browser though, right?

exogen: Correct.

PR #66, opened 2017-09-14
extending htz-react-server with graphql server

eliagrady:
Not yet working


TxHawks:
Can you rename the PR?  It's a duplicate of the previous PR

packages/apps/haaretz.co.il/lib/createClient.js
```diff
@@ -1,91 +1,10 @@
...
-          operationName
-        );
-      }
-      catch (err) {
-        return Promise.reject(err);
-      }
-    },
-  };
-}
-
```
exogen: Because context is only used by the resolvers, and the resolvers are executed on the server now, the client has no need to create all this stuff.

packages/libs/htz-react-server/src/createContext.js
```diff
@@ -0,0 +1,24 @@
+import DataLoader from 'dataloader';
+
+export function createLoaders() {
+  // By default, `DataLoader` just caches the results forever, but we should
+  // eventually expunge them from the cache.
+  const pageLoader = new DataLoader(keys =>
+    Promise.all(
+      keys.map(path =>
+        fetch(`http://elia.haaretz.co.il:8080/papi${path}`).then(response =>
```
exogen: `path` should already start with a `/` here, so don'y duplicate it in the URL.

packages/libs/htz-react-server/src/createContext.js
```diff
@@ -0,0 +1,24 @@
...
+          response.json()
+        )
+      )
+    )
+  );
+  return { pageLoader, };
+}
+
+export default function createContext(cookies) {
+  const loaders = createLoaders();
```
exogen: See [my comment](https://github.com/Haaretz/htz-frontend/issues/18#issuecomment-329436614) about potentially reusing the `pageLoader` in the future.

packages/libs/htz-react-server/src/index.js
```diff
@@ -0,0 +1,65 @@
...
+const handler = app.getRequestHandler();
+
+app
+  .prepare()
+  .then(() => {
+    const server = express();
+    server.use(compression()); // Compress responses.
+    server.use(helmet()); // Various security-minded settings.
+
+    server.use(
```
exogen: The order here is important. In @eliagrady's previous commits, the `*` and `/.*` handlers were moved before `/graphql` and `/graphiql`, meaning they handled the requests. More specific URLs should go first so that other handlers don't match them.

exogen:
I got this into a place where I'm getting:

> FetchError: request to http://elia.haaretz.co.il:8080/papi/ failed, reason: getaddrinfo ENOTFOUND elia.haaretz.co.il elia.haaretz.co.il:8080

Which is what I would expect since it appears I don't have access to that endpoint.

Not 100% sure it's working, but it's working up to that point! See my comments for the changes.

packages/libs/htz-react-server/src/index.js
```diff
@@ -21,6 +26,25 @@ app
...
+      })
+    );
+
+    server.use(
+      '/graphiql',
+      graphiqlExpress({
+        endpointURL: '/graphql',
+      })
+    );
+
```
exogen: Re-adding this comment so GitHub doesn't hide it:

The order here is important. Since the `*` and `/.*` handlers below would also match `/graphql` and `/graphiql`, these need to come first.

exogen:
Another tip: since there's no `dev` or `start` command in `htz-react-server` itself yet, this all needs to be tested in the actual `haaretz.co.il` test. When you make changes in `htz-react-server`, don't forget to `yarn run build`, otherwise the app won't be running the latest server code and you'll wonder why stuff isn't working. The dev experience here needs to be improved (some kind of watch mode).

packages/apps/haaretz.co.il/lib/createClient.js
eliagrady: Cool, so I only write resolver logic once, in the htz-react-server right?

packages/libs/htz-react-server/src/createContext.js
eliagrady: is `path` a relative path? (i.e: /news/1.323) ?

packages/libs/htz-react-server/src/createContext.js
eliagrady: For simplicity, I think we can keep using DataLoader as a simple map - and have the advanced LRU and time expiry logic baked in at a later stage

eliagrady:
Now we're getting somewhere! /graphiql and /graphql endpoints are working, but using the graphiql GUI I cannot get a query for a page to work:

```
{
  page("/") {
  contentName
  inputTemplate
  }
}
```
I'm getting

`{
  "errors": [
    {
      "message": "Syntax Error GraphQL request (2:8) Expected Name, found String \"/papi\"\n\n1: {\n2:   page(\"/papi\") {\n          ^\n3:   contentName\n",
      "locations": [
        {
          "line": 2,
          "column": 8
        }
      ]
    }
  ]
}`

so I must be writing the queries all wrong

exogen:
Awesome! You just need to specify the field argument name, it's required unlike in JS functions.

```
{
  page(path: "/") {
  contentName
  inputTemplate
  }
}
```
(the `/papi` part should already be part of the `fetch`, the `path` argument in the resolver is the input argument value here!)

packages/libs/htz-react-server/src/createContext.js
exogen: It's whatever you pass into the field argument, e.g.: `page(path: "/new/1.323")`

packages/apps/haaretz.co.il/lib/createClient.js
exogen: Exactly!

eliagrady:
I've added basic access to /json/cmlink/{CommentElementContentId} data endpoint with it's respected resolver. 
For example, a query that looks like this:
```
{
  comments(contentId:"7.1544") {
    comments {
      author
      articleId
      parentCommentId
      relatedComments
      publishingDateSortable
      title
      commentText
      publishingDateForDisplay
      publishingDateIST
      publishingDateForDisplayMobile
      reviewState
      isEditorPick
      publishingDateForData
      publishingDateForDisplayApp
    }
    properties
  }
}
```
would yield:
 ```
{
  "data": {
    "comments": [...],
    "properties": {...}
  }
}
```
when comments is our main comment data, and properties contains metadata.

File: .gitignore [hunk 6]
```diff
@@ -1,3 +1,9 @@
+# IDE specific metadata
+.idea/
+
+# Yarn specific
+.yarnclean
```
TxHawks: With yarn >= `1.0.0` we no longer need the `.yarnclean` file. It can just be removed

packages/libs/htz-react-server/src/createContext.js
```diff
@@ -2,7 +2,7 @@ import DataLoader from 'dataloader';
 
 export function createLoaders() {
   // By default, `DataLoader` just caches the results forever, but we should
-  // eventually expunge them from the cache.
+  // <TODO>eventually expunge them from the cache.</TODO>
```
TxHawks: Why `<TODO></TODO>` and not `TODO:` or `FIXME:`?

packages/libs/htz-react-server/src/createContext.js
```diff
@@ -12,7 +12,16 @@ export function createLoaders() {
       )
     )
   );
-  return { pageLoader, };
+  const cmlinkLoader = new DataLoader(keys =>
+    Promise.all(
+      keys.map(contentId =>
+        fetch(`http://elia.haaretz.co.il:8080/json/cmlink/${contentId}`).then(response =>
```
TxHawks: I'd say add a `TODO` here to remove hardcoded url so that the linter nags about it (It's a warning, not an error)

TxHawks:
@eliagrady, Unless you're planning to squash these together later, can you amend the last commit so that the message says something about what it does rather that what it is? "merge guy" isn't really helpful in understanding what the code changes are.

Issue #82, opened 2017-12-03
Article body - embeds


adibi83:
## Issue Type

* [ ] Bug fix
* [X] New Feature
* [ ] Update to existing feature
* [ ] General discussion

## Description

Implementation of new embed elements in new article page:
* [ ] Global `<Embed />` componet
* [ ] Arti-Media video
* [ ] 103FM Audio
* [ ] BandCamp
* [ ] Bloomberg Video
* [ ] CNN video
* [ ] Facebook
* [ ] Facebook comments
* File upload
  * [ ] PDF 
  * [ ] Word 
  * [ ] Excel 
  * [ ] PowerPoint
* [ ] Giphy
* [ ] Google maps
* [ ] Guardian video
* [ ] Instagram
* [ ] NYT videos
* [ ] Omny-studio
* [ ] Pinterest
* [ ] PlayBuzz
* [ ] Reuters video
* [ ] SoundCloud
* [ ] Tline
* Twitter
  * [ ] Tweet
  * [ ] Twitter video
  * [ ] Twitter timeline
* [ ] Vimoe
* [ ] Waze map
* [ ] YouTube 
* [ ] legacy embeds

### Configuration options

Same as existing elements in velocity implementation

### ~Measurement Requirements~

~Not relevant~
  

### A11Y Concerns

**Assistive tech (scrren readers):**
  * Announce embed type one step before it is navigated to, offer to skip it

## High Level Design
**Participants:**
  1. @EranM6 
  2. @TxHawks

**Overview:**
  * The global `<Embed />` component is the public API for all embed types, and the only one directly exported by the package. The actual embed used and its content will be determined by the props passed to `<Embed />`, where all logic will be implemented and abstracted away from the consumer.
  * **API Surface (props):**
    * `inputTemplate` - what this?

## Testing Requirements

  * Error validation
  * snapshot testing to validate rendering

## Work Time Estimation

12 hours

PR #108, opened 2017-12-21
Feature/embeds

EranM6:
**Related issue(s):** #82

## Description
3rd party embedded content components

**Notes:**
More tests needs to be written once Next 4.3.0 is release, and testing `dynamic()` is made possible.

## Checklist

Usability:
  * [x] Developer
  * [ ] Project manager

Tests
  
  * [ ] UI snapshot testing 
  * [ ] Unit testing
  * [ ] Integration testing
  * [ ] E2E testing


packages/components/htz-components/src/components/Embed/elements/ArtiMedia.js
```diff
@@ -0,0 +1,114 @@
...
+  return {
+    margin: '0',
+    paddingBottom: aspect,
+    height: '0',
+    overflow: 'hidden',
+    position: 'relative',
+  };
+};
+
+const VideoWrapper = createComponent(videoWrapper, 'figure');
```
TxHawks: This and the style function above is duplicated in a few places. Extract them to their own component and import it in the various call sites.

TxHawks: The `VideoWrapper` should be a `div`, not a figure. The `figure` should wrap the embed and the caption together, so should definitely not have an aspect ratio.

packages/components/htz-components/src/components/Embed/elements/ArtiMedia.js
```diff
@@ -0,0 +1,114 @@
...
+      <div>
+        <VideoWrapper aspectRatio={'16/9'}>
+          <VideoElement id={playerId} className="arti-media-video" />
+        </VideoWrapper>
+        <Caption
+          caption={this.props.caption}
+          credit={this.props.credit}
+        />
+      </div>
+    );
```
TxHawks: See my previous comments about `<Caption />`. I'd go for a return value of:
```jsx
<VideoWrapper aspectRatio={'16/9'}>
  <VideoElement id={playerId} className="arti-media-video" />
</VideoWrapper>
```

packages/components/htz-components/src/components/Embed/elements/ArtiMedia.js
```diff
@@ -0,0 +1,114 @@
...
+  }
+
+  render() {
+    const settings = this.props.settings;
+    const playerId = settings.playerId;
+
+    return (
+      <div>
+        <VideoWrapper aspectRatio={'16/9'}>
+          <VideoElement id={playerId} className="arti-media-video" />
```
TxHawks: Is the `className` needed?

Issue #85, opened 2017-12-03
Article body - Quote component


adibi83:
## Issue Type

* [ ] Bug fix
* [X] New Feature
* [ ] Update to existing feature
* [ ] General discussion

## Related Issues
[#0]

## Description

* [ ] Regular quote
* [ ] Quotation
* [ ] Quote with image

## Configuration
[רכיב quote.docx](https://github.com/Haaretz/htz-frontend/files/1524833/quote.docx)

### Measurement Requirements
  * [ ] ds - clicks
  

### SEO Requirements

  * [ Insert requirement ]

### A11Y Concerns

**Focus management:**
  * [ Insert requirement ]

**Assistive tech (scrren readers):**
  * [ Insert requirement ]

**Color contrast:**
  * [ Insert requirement ]

## Links

**Designs:**
  * Desktop: https://app.zeplin.io/project/59b4e63a7a92547269089328/screen/59b4e670b310f781615188dd 
  * Mobile: 

## High Level Design
**Participants:**
  1. [ Insert name ]

**Overview:**
  * [ Insert description ]

## Testing Requirements

  * [ Insert description ]

## Work Time Estimation

[n] hours

PR #127, opened 2018-01-22
Feature/quote

EranM6:
**Related issue(s):** #85

## Description

**Notes:**
Missing image, still waiting for the Image cmponent.

## Checklist

Design:
  * [ ] Approved by designer

Usability:
  * [x] Developer
  * [ ] Project manager

Tests
  
  * [x] UI snapshot testing


packages/components/htz-components/src/components/Quote/Quote.js
```diff
@@ -0,0 +1,136 @@
...
+   */
+  credit: PropTypes.string,
+  /**
+   * Quotes position, **only** in Magazine article (mid-left, mid-right, ect).
+   */
+  position: PropTypes.string,
+  /**
+   * After which paragraph should the Quote be displayed (for Magazine article).
+   */
+  afterParagraph: PropTypes.oneOfType([
```
TxHawks: I'm not sure I understand the purpose of this prop... Isn't the position controlled by the parent component? Also can't see how it is used in the code

packages/components/htz-components/src/components/Quote/Quote.js
```diff
@@ -0,0 +1,136 @@
...
+  marginBottom: '1rem',
+});
+const TopBorder = createComponent(topBorderStyle, 'span');
+
+const getStyleObj = (theme, quoteType) => {
+  switch (quoteType) {
+    case 'image' :
+      return {
+        ':before': {
+          content: "'\"'",
```
TxHawks: ?

packages/components/htz-components/src/components/Quote/Quote.js
```diff
@@ -0,0 +1,136 @@
...
+const topBorderStyle = ({ theme, }) => ({
+  borderTopWidth: '1rem',
+  borderTopColor: theme.color('primary'),
+  borderTopStyle: 'solid',
+  width: '16.5rem',
+  height: '1rem',
+  display: 'inline-block',
+  marginBottom: '1rem',
+});
+const TopBorder = createComponent(topBorderStyle, 'span');
```
TxHawks: Not a big deal, but I'd use a `::before` pseudo element for the border.

TxHawks:
Eran, can you show this to max and Adi as well (visually I mean, in the styleguide)

packages/components/htz-components/src/components/Quote/Quote.js
EranM6: not always. in Magazine article its position is set in the element's meta

packages/components/htz-components/src/components/Quote/Quote.js
EranM6: according to the design, when there's an image, the quoted sentence shuold be surrounded by a quotation marks

packages/components/htz-components/src/components/Quote/Quote.js
EranM6: It's much easier to control its alignment that way.
And beside that, beause this kind of decoration repeats itself in our sites (at least the current ones), perhaps it shuold be a component of its own ??

packages/components/htz-components/src/components/Quote/Quote.js
TxHawks: Let's drop this. This isn't something that we'll use in magazine articles either

Issue #111, opened 2018-01-09
Article Body - Mobile Quick Registration Component


adibi83:
## Issue Type
**New Feature**

## Description
Quick registration element (available only in mobile).
Polopoly input template: 14.1524 (com.polobase.quickNewsletterRegistration)

### Configuration options

  * [ ] config is set in "HTML elements" tab on 14.901 (com.tm.articleViewWebMobile)

### Measurement Requirements
  
  * ds 
    * [ ] [ Insert requirements ]
  * tag manager
    * [ ] [ Insert requirement ]
  

### A11Y Concerns

**Focus management:**
  * [ ] [ Insert requirement ]

**Assistive tech (scrren readers):**
  * [ ] [ Insert requirement ]

**Color contrast:**
  * [ ] [ Insert requirement ]

## Links

**Designs:**
https://app.zeplin.io/project/59b4e63a7a92547269089328/screen/5a549b045cc3742d090887b3?

## High Level Design
**Participants:**
  1. [ Insert name ]

**Overview:**
  * [ Insert description ]

## Testing Requirements

  * [ Insert description ]

## Work Time Estimation

[n] hours

PR #128, opened 2018-01-23
feat(htz-components): mobile quick registration

EranM6:
**Related issue(s):**#111

## Description

**Notes:**

## Checklist

Design:
  * [ ] Approved by designer

Usability:
  * [x] Developer
  * [ ] Project manager

Tests
  
  * [ ] UI snapshot testing


packages/components/htz-components/src/components/MobileQuickRegistration/MobileQuickRegistration.js
```diff
@@ -0,0 +1,98 @@
...
+
+const propTypes = {
+  /**
+   * Element's main text, displayed above the button.
+   */
+  teaserBody: PropTypes.string.isRequired,
+  /**
+   * doptint || doptinf
+   */
+  doubleOptIn: PropTypes.string.isRequired,
```
TxHawks: ???
Is this a tuple and these are the two options that it can take? If so, it should be `PropTyoes.oneOf(['doptin', 'doptinf'])`, and then, can't we just get a boolean here?

packages/components/htz-components/src/components/MobileQuickRegistration/MobileQuickRegistration.js
```diff
@@ -0,0 +1,98 @@
...
+   * Button's inner text.
+   */
+  teaserButton: PropTypes.string.isRequired,
+};
+
+const wrapperStyle = ({ theme, }) => ({
+  position: 'relative',
+  textAlign: 'center',
+  borderTopStyle: 'solid',
+  borderTopWidth: '2px',
```
TxHawks: Border and padding should be set up using border functions from htz-css-tools, so that it doesn't break the vertical rhythm

packages/components/htz-components/src/components/MobileQuickRegistration/MobileQuickRegistration.js
```diff
@@ -0,0 +1,98 @@
...
+      <ButtonText>{teaserButton}</ButtonText>
+    </Button>;
+
+  return (
+    <RegistrationWrapper>
+      <IconWrapper>
+        <IconMailAlert size={5} color='primary' />
+      </IconWrapper>
+      <TeaserBody>{teaserBody}</TeaserBody>
+      <Link href={href} content={buttonContent}/>
```
TxHawks: Should `Link` be used for `mailto` links? I thought it only makes sense in t g e context of routing

TxHawks: Also, it is invalid html to put one interactive element (`button`) inside another (`a`)

PR #129, opened 2018-01-23
feat(htz-components): article's tags element

EranM6:

ISSUES CLOSED: #89

**Related issue(s):** #0

## Description

**Notes:**

## Checklist

Design:
  * [ ] Approved by designer

Usability:
  * [ ] Developer
  * [ ] Project manager

Tests
  
  * [x] UI snapshot testing


packages/components/htz-components/src/components/Tags/Tags.js
```diff
@@ -0,0 +1,63 @@
...
+       * Tag's name to be displayed.
+       */
+      contentName: PropTypes.string.isRequired,
+    })
+  ).isRequired,
+};
+
+const wrapperStyle = ({ theme, }) => ({
+  ...(theme.type(-1)),
+  '& > li': {
```
TxHawks: Why not add these styles to the <Tag /> component? Using `& > <tag>` should really only be used as last resort.
It's creates styles that are:
1) brittle
2) have higher specificity, which is especially bad for atomic CSS
3) not very reusable.

packages/components/htz-components/src/components/Tags/Tags.js
```diff
@@ -0,0 +1,63 @@
...
+    display: 'inline',
+    ':not(:last-child)': {
```
TxHawks: For the same reasons mentioned above, I'd do this as a prop to a `<Tag />` component

packages/components/htz-components/src/components/Tags/Tags.js
```diff
@@ -0,0 +1,76 @@
...
+       */
+      contentName: PropTypes.string.isRequired,
+    })
+  ).isRequired,
+};
+
+const wrapperStyle = ({ theme, }) => ({
+  ...(theme.type(-1)),
+});
+const TagsWrapper = createComponent(wrapperStyle, 'ul');
```
TxHawks: This can probably just be named `Tags`, and then we'd have `Tags -> Tag`, which reads more clearly.
I'd expect `TagsWrapper` to be an extra component wrapping `Tags` for styling or logic

packages/components/htz-components/src/components/Tags/Tags.js
```diff
@@ -0,0 +1,76 @@
...
+
+const contentStyle = ({ theme, }) => ({
+  color: theme.color('neutral', '-2'),
+  textDecoration: 'underline',
+});
+const Content = createComponent(contentStyle, 'span');
+
+const tagStyle = ({ theme, }) => ({
+  display: 'inline',
+  ':not(:last-child)': {
```
TxHawks: This increases specificity by 2, which isn't ideal.

I think doing this with a `hasMargin` prop that is passed by the parent through `Children.map` is a better approach.

Another option is to add the following to the parent:

```
'& > *  + *': { marginStart: '1rem`, }
```

This solution is nice because it removes the need for an iterator and because it uses the universal selector (`*`) it doesn't increase the specificity of the rule

packages/components/htz-components/src/components/Tags/Tags.js
```diff
@@ -0,0 +1,76 @@
...
+
+/**
+ * Returns the article related tags.
+ */
+function Tags({ tagsList, }) {
+  return(
+    <TagsWrapper>
+      <Prefix>{i18n.tagsElement.prefix}</Prefix>
+      {tagsList.map((tag, i) =>
+        <Tag key={i}>
```
TxHawks: Using indices as keys is a bad idea:
https://medium.com/@robinpokorny/index-as-a-key-is-an-anti-pattern-e0349aece318

This is actually the same as not passing a key at all, since this is what React falls back to under the hood when you don't provide a key.

Using the `tag.contentName` is a good option

packages/components/htz-components/src/components/Tags/Tags.js
```diff
@@ -0,0 +1,76 @@
...
+  ...(theme.type(-1)),
+});
+const TagsWrapper = createComponent(wrapperStyle, 'ul');
+
+const prefixStyle = ({ theme, }) => ({
+  fontWeight: '700',
+  color: theme.color('secondary'),
+  marginEnd: '1rem',
+});
+const Prefix = createComponent(prefixStyle, 'span');
```
TxHawks: `TagPrefix`?

packages/components/htz-components/src/components/Tags/Tags.js
```diff
@@ -0,0 +1,76 @@
...
+
+const contentStyle = ({ theme, }) => ({
+  color: theme.color('neutral', '-2'),
+  textDecoration: 'underline',
+});
+const Content = createComponent(contentStyle, 'span');
```
TxHawks: `TagContent`?

packages/components/htz-components/src/components/Tags/Tags.js
```diff
@@ -0,0 +1,76 @@
...
+  }
+});
+const Tag = createComponent(tagStyle, 'li');
+
+/**
+ * Returns the article related tags.
+ */
+function Tags({ tagsList, }) {
+  return(
+    <TagsWrapper>
```
TxHawks: `Prefix` shouldn't be placed inside the `ul`, it should be an `h4` placed before it.

packages/components/htz-components/src/components/Tags/Tags.js
TxHawks: Actually, it should probably be `TagsTitle`

packages/components/htz-components/src/components/Tags/Tags.js
TxHawks: Ok, sorry, scrap my previous comment (GitHub's stupid movie interface won't let me edit it).
It should just be renamed `TagList`

packages/components/htz-components/src/components/Tags/Tags.js
```diff
@@ -0,0 +1,76 @@
...
+  ':not(:last-child)': {
+    marginEnd: '1rem',
+  }
+});
+const Tag = createComponent(tagStyle, 'li');
+
+/**
+ * Returns the article related tags.
+ */
+function Tags({ tagsList, }) {
```
TxHawks: Making `Tags` a flexbox container makes it trivial to place the title next to the ul

packages/components/htz-components/src/components/Tags/Tags.md
```diff
@@ -0,0 +1,30 @@
+<!-- START doctoc generated TOC please keep comment here to allow auto update -->
+<!-- DON'T EDIT THIS SECTION, INSTEAD RE-RUN doctoc TO UPDATE -->
+**Table of Contents**
+
+- [Tags examples](#tags-examples)
+
+<!-- END doctoc generated TOC please keep comment here to allow auto update -->
+
+### Tags example
```
TxHawks: Can you add an explanation of what this component is used for?

PR #136, opened 2018-01-24
Feature/links block

EranM6:
**Related issue(s):** #0

## Description

**Notes:**
Some of the desing is still missing ('display more' button and current article in the list)

## Checklist

Design:
  * [ ] Approved by designer

Usability:
  * [ ] Developer
  * [ ] Project manager

Tests
  
  * [x] UI snapshot testing


packages/components/htz-components/src/components/RelatedArticles/LinksBlock.js
```diff
@@ -0,0 +1,118 @@
...
+import PropTypes from 'prop-types';
+import { createComponent, } from 'react-fela';
+import { parseComponentProp, } from '@haaretz/htz-css-tools';
+import ListForSeries from "./series components/listForSeries";
+
+const propTypes = {
+  /**
+   * The name of the series to display.
+   */
+  seriesTitle: PropTypes.string.isRequired,
```
TxHawks: Is this not a series though, right?

EranM6: It is.
The props for this one and Series articles are the same, except that in this one the articles comes with the Auothor

PR #138, opened 2018-01-24
fix(htz-components): fixed minor bugs in Caption component

EranM6:

fixed some bus in the Caption components, such as:
* console log error
* background color didn't work
* missing test

**Related issue(s):** #0

## Description

**Notes:**

## Checklist

Design:
  * [ ] Approved by designer

Usability:
  * [ ] Developer
  * [ ] Project manager

Tests
  
  * [x] UI snapshot testing


packages/components/htz-components/src/components/Caption/Caption.js
```diff
@@ -13,19 +13,19 @@ const setColor = (prop, value, getColor) => {
 
 const captionWrapper = ({
   theme,
-  bgc,
+  backgroundColor,
   color,
   typeStyles, // Not responsive. can only take "number"
```
TxHawks: Why?
I know this isn't part of the changes in this commit,  but it should be made possible to take a responsive array

EranM6: `backgroundColor` is the name of the prop (line 183), `bgc` is the parameter's name that may come from the Theme (That's why the backgrond color doesn't work correctly right now)

PR #158, opened 2018-02-05
Feature/add scroll components

kedarguy:
**Related issue(s):** #0

**Notes:**

Added an example of how to use in app, need to decide if we want to leave it.

I think that because its not trivial and currently docs are not great for Apollo-Link-State we should leave a working example that includes building a component in htz-components and importing and using it in haaretz.co.il

Tests
  * [x] UI snapshot testing


packages/components/htz-components/src/components/CommentList/__snapshots__/CommentList.test.js.snap
```diff
@@ -410,7 +410,7 @@ Object {
...
...nt:}.htz-bi:after{height:1px}.htz-bj:after{position:absolute}.htz-bk:after{width:100%}.htz-bl:after{bottom:0rem}.htz-bm{align-items:flex-start;-webkit-box-align:start;-webkit-align-items:flex-start}.htz-bn{margin-left:4rem}.htz-bo{align-items:center;-webkit-box-align:center;-webkit-align-items:center}.htz-bp{margin-right:2rem}.htz-bq{cursor:pointer}.htz-br{vertical-align:center}.htz-bv{border:none}.htz-bw{color:#AB353B}.htz-bx{border-bottom-color:#DAE9F2}.htz-by{color:#B4B4B4}.htz-bz{font-style:italic}@media (max-width: 80em){.htz-n{font-size:4rem}.htz-o{line-height:1.25em}.htz-w{font-size:2.6666666666666665rem}.htz-x{line-height:1.5em}.htz-ab{font-size:2.3333333333333335rem}.htz-ac{line-height:1.2857142857142858em}.htz-ap{font-size:1.8333333333333333rem}.htz-aq{line-height:1.6363636363636365em}.htz-bs{font-size:2rem}}@media (min-width: 80em){.htz-p{font-size:3.857142857142857rem}.htz-q{line-height:1.2962962962962963em}.htz-y{font-size:2.5714285714285716rem}.htz-z{line-height:1.5555555555555556em}.htz-ae{font-size:2.2857142857142856rem}.htz-af{line-height:1.3125em}.htz-ar{font-size:1.7142857142857142rem}.htz-as{line-height:1.75em}.htz-bt{font-size:2rem}.htz-bu{line-height:1.5em}}",
```
kedarguy: check :before{content: \\"\\"} changed to before:content:{}

PR #189, opened 2018-02-26
Feature/footer

TawKhoury:
Footer component for desktop and mobile

## Checklist

Design:
  * [ ] Approved by designer

Usability:
  * [x] Developer
  * [ ] Project manager

Tests
  
  * [x] UI snapshot testing 
  * [x] Unit testing


packages/components/htz-components/src/components/Footer/Footer.js
```diff
@@ -0,0 +1,22 @@
...
+const Footer = ({ theme, }) => (
+  <Media query={`(max-width: ${theme.bps.widths.s}px)`}>
+    {matches =>
+      (matches ? (
+        <MobileView footerMobileListsI18n={theme.footerMobileListsI18n} />
+      ) : (
+        <DesktopViewWithApollo theme={theme} />
+      ))
+    }
+  </Media>
```
TxHawks: I've made a `<Media />` component that uses our own API for breakpoints, please use that.

Also, note that you render a component on SSR, and then switch it if the query doesn't match. This is wasteful. Since the footer isn't a critical above-the-fold component, it's okay to defer its rendering to the front end, and only render the correct one there.

PR #211, opened 2018-03-08
fix(htz-components): textInput, Note and Button Components minor fix

TawKhoury:

* passed variant to Note in TextInput component 
* more code refactor in Note.js (dynamically extend color attr using parseComponentProp )


packages/components/htz-components/src/components/Button/Button.js
```diff
@@ -18,6 +18,7 @@ const btnVariants = PropTypes.oneOf([
   'primaryOpaque',
   'negative',
   'negativeOpaque',
+  'newsletter',
```
TxHawks: Newsletter isn't a good name for a button variant because it's very specific and means it will either not be reused but rather duplicated, or worse - be reused in places where the naming is inappropriate.  Also, I don't see any change in the theme that coresponds.

I'd call this button variant `neutral`.

packages/components/htz-components/src/components/TextInput/textInputVariantType.js
```diff
@@ -1,3 +1,7 @@
 import { oneOf, } from 'prop-types';
 
-export const textInputVariantType = oneOf([ 'primary', 'primaryInverse', ]);
+export const textInputVariantType = oneOf([
+  'primary',
+  'primaryInverse',
+  'newsletter',
```
TxHawks: See my other comment on variant naming

packages/components/htz-components/src/components/Button/Button.js
TawKhoury: Thanks!

packages/components/htz-components/src/components/TextInput/textInputVariantType.js
TawKhoury: Done

TxHawks:
This needs to be rebased against master

packages/components/htz-components/src/components/Note/Note.js
```diff
@@ -45,12 +50,27 @@ const defaultProps = {
...
+  ],
 });
 
+function setVariant(prop, variant, getColor, isError) {
+  return {
+    color: isError
+      ? getColor('input', `${variant}ErrorTextNote`)
+      : getColor('input', `${variant}TextNote`),
+  };
+}
```
TxHawks: This isn't a variant - a variant is a useful abstraction only when you actually need to make a set of palette changes, and name them semantically. For example, in Button, a variant changes the `color`, `background-color`, `border-color` and their sate.

Here it's a meaningless abstraction, that only adds cognitive overhead, since all you are trying to do is change the `color`

PR #234, opened 2018-03-27
fix: dfp injector fix

eliagrady:
affects: @haaretz/haaretz.co.il, @haaretz/htz-components, @haaretz/app-utils, @haaretz/dfp,
@haaretz/htz-react-server

**Table of Contents**

- [Description](#description)
- [Checklist](#checklist)

**Related issue(s):** #0

## Description

**Notes:**

## Checklist

Design:
  * [ ] Approved by designer

Usability:
  * [ ] Developer
  * [ ] Project manager

Tests
  
  * [ ] UI snapshot testing 
  * [ ] Unit testing
  * [ ] Integration testing
  * [ ] E2E testing


packages/libs/app-utils/src/apollo/withData.js
```diff
@@ -86,7 +87,7 @@ export default Component => {
             // it's gone now. For now, just throw a 404 if any error message is
             // 'Not Found', and a 500 otherwise.
             console.error('getDataFromTree() failed!');
-            // console.error(err);
+            console.error(err);
```
eliagrady: We should display a better message for errors for as long we do not have a stable build

packages/libs/htz-react-server/src/index.js
```diff
@@ -88,7 +88,14 @@ app
     // /1\.\d+.*$/
     // /^(\/.*(?!\d\.\d+))$/
     server.get([ /^.*(1\.\d+){1}$/, ], (req, res) => {
-      console.log('captured an article!');
+      if (DEV) {
```
eliagrady: for now, while working on DEV, I want to explicitly state the difference between req.path which can be /.premium-LIVE-1.234 and req.params[0] which is in this example should be 1.234

packages/libs/htz-react-server/src/index.js
```diff
@@ -88,7 +88,14 @@ app
...
+        console.log(
+          'captured an article at  req.path: ',
+          req.path,
+          ' req.params[0]: ',
+          req.params[0]
+        );
+      }
       if (!req.params[0].startsWith('/')) {
```
eliagrady: As per previous comment, 'req.params[0]' has no leading path slash, which graphql resolver currently expects.

packages/apps/haaretz.co.il/layouts/MainLayout.js
```diff
@@ -128,7 +129,9 @@ export class MainLayout extends React.Component {
         <UserInjector />
         {LoginExample}
         {RegisterExample}
-        {/* <DfpInjector /> */}
+        <NoSSR>
```
eliagrady: Having `<DfpInjector />` as a `<NoSSR>` rendered element makes sense for ads to be rendered on client only.

packages/components/htz-components/src/components/Ads/DfpInjector.js
```diff
@@ -129,7 +130,7 @@ DfpInjector.defaultProps = defaultProps;
 
 export default graphql(GET_AD_MANAGER, {
   options: props => ({
-    variables: { path: '/', },
+    variables: { path: window.location.path, },
```
eliagrady: While `<DfpInjector />` is rendered using `<NoSSR>`, it is safe to use the `window` global.

packages/apps/haaretz.co.il/layouts/MainLayout.js
TxHawks: It shouldn't be the consumer's responsibility to k ow to use `<NoSSR>` to wrap the injector. The exported injector should already be wrapped with in internally, as in:

```jsx
export default () => (<NoSSR><DfpInjector /><NoSSR>)
```

packages/apps/haaretz.co.il/layouts/MainLayout.js
```diff
@@ -27,6 +25,9 @@ const ScrollInjector = dynamic(
   import('../components/ScrollListener/ScrollListener'),
   { ssr: false, }
 );
+const DfpInjector = dynamic(import('../components/Dfp/DfpInjector'), {
```
eliagrady: Dynamic import make sure we only load Dfp Context at client

packages/components/htz-components/src/components/Ads/DfpInjector.js
```diff
@@ -52,7 +52,7 @@ const initDfpScript = (dfpConfig = {}) => {
 
 export const GET_AD_MANAGER = gql`
   query getDfpConfig($path: String!) {
-    page(path: $path) @client {
+    page(path: $path) {
```
eliagrady: If the page is cached, then it's data will be used (from client cache) - otherwise it will fetch the data from the server.
In either case, we are safer using non-client query.

PR #391, opened 2018-05-09
Refactor/design touch ups 2

kedarguy:


packages/apps/purchase-page/theme/consts/i18n.js
```diff
@@ -215,6 +215,7 @@ export const stage3 = Object.freeze({
     }),
     email: Object.freeze({
       errorText: 'נא למלא כתובת דוא"ל',
+      errorTextLegalEmail: 'יש להזין כתובת דוא"ל תקנית',
```
TxHawks: Why is this called legal email?

kedarguy: will change to valid

PR #508, opened 2018-06-19
feat(no-ssr): adding no ssr component

TawKhoury:


packages/components/htz-components/src/components/NoSSR/NoSSR.js
```diff
@@ -0,0 +1,37 @@
...
+     */
+    children: PropTypes.node,
+    /**
+     * Trigger a function on server side rendering.
+     */
+    onSSR: PropTypes.func,
+  };
+  static defaultProps = {
+    children: null,
+    onSSR: () => {},
```
TxHawks: This should be `null`

packages/components/htz-components/src/components/NoSSR/NoSSR.js
```diff
@@ -0,0 +1,37 @@
...
+  static propTypes = {
+    /**
+     * Nodes rendered inside `NoSSR`.
+     * Passed to the underlying react element
+     */
+    children: PropTypes.node,
+    /**
+     * Trigger a function on server side rendering.
+     */
+    onSSR: PropTypes.func,
```
TxHawks: Why is this a function? Isn't the idea to render something on the server? 

Shouldn't it be `PropTypes.node`?

TxHawks:
@TawKhoury I just noticed that this is missing documentation. Can you add this in another PR?

PR #562, opened 2018-07-03
feat: detection for next-supported urls

eliagrady:
affects: @haaretz/htz-components, @haaretz/app-utils

**Table of Contents**

- [Description](#description)
- [Checklist](#checklist)

**Related issue(s):** #0

## Description

**Notes:**

## Checklist

Design:
  * [ ] Approved by designer

Usability:
  * [ ] Developer
  * [ ] Project manager

Tests
  
  * [ ] UI snapshot testing 
  * [ ] Unit testing
  * [ ] Integration testing
  * [ ] E2E testing


packages/components/htz-components/src/components/HtzLink/__tests__/isNextLink.test.js
```diff
@@ -0,0 +1,103 @@
...
+  ],
+];
+
+describe('isNextLink', () => {
+  describe('single test urls', () => {
+    const urlTest = testUrls[0];
+    it(`isNextLink ${urlTest[0]} should classify correctly`, () => {
+      expect(isNextLink(urlTest[0])).toEqual(urlTest[1]);
+    });
+  });
```
TxHawks: This `describe` block redundent no? It's covered by the loop in the following block

packages/components/htz-components/src/components/HtzLink/isNextLink.js
```diff
@@ -1,26 +1,43 @@
+/* eslint-disable no-unused-vars */
 import config from 'config';
-// import { breakUrl, } from '@haaretz/app-utils';
+import { breakUrl, } from '@haaretz/app-utils';
 
-const domain = config.has('domain') && config.get('domain');
-const isLocal = new RegExp(`(^((?!((https?:)?\\/\\/)).)*$)|(${domain})`);
+// const domain = config.has('domain') && config.get('domain');
+// const isLocal = new RegExp(`(^((?!((https?:)?\\/\\/)).)*$)|(${domain})`);
```
TxHawks: Any point in keeping these lines? Let's erase them.

packages/components/htz-components/src/components/HtzLink/__tests__/isNextLink.test.js
eliagrady: It is, but it allows up to quickly get a specific url from the test suite and work on it instead of needing to write it. The error message for failed tests is not that great.
We will be adding more test cases soon, and frequently.
could be a comment as well.

packages/components/htz-components/src/components/HtzLink/isNextLink.js
eliagrady: agreed. should be cleaned.

Issue #573, opened 2018-07-04
Add ILX DIVs


adibi83:
## Issue Type

**Update to existing feature**

## Description
Add ILX DIV (<div class="FI_PH"></div> ) before the footer and outside the page's layout

PR #599, opened 2018-07-11
feat(first impression placeholders): add firstimpression place holder…

kedarguy:
… to top of footer and outside viewport
remove placeholder from last element of article body

ISSUES CLOSED: #573


packages/components/htz-components/src/components/Footer/Footer.js
```diff
@@ -231,6 +233,9 @@ class Footer extends React.Component {
         <GoogleReMarketingTag />
         <ChartBeat />
         <ChromePush />
+        <FelaComponent style={{ position: 'absolute', right: '-1000rem', }}>
```
TxHawks: What's the purpose of this?

kedarguy: Removed Wrapping FelaComponent

PR #615, opened 2018-07-16
feat(styleguide): add action buttons mock data

tamirbenCA:


packages/components/htz-components/styleguide/ApolloMockClient.js
```diff
@@ -53,6 +53,8 @@ const stateLink = withClientState({
     a11yToggle: false,
     articleId: '1.2345',
     commentsElementId: '1.2345',
+    platform: 'web',
```
TxHawks: I think this should be dynamically set based on the screen size, otherwise we can't test mobile. @TawKhoury, am I right?

PR #624, opened 2018-07-17
fix: add ASSET_PREFIX config variable fix static fonts

eliagrady:
affects: @haaretz/haaretz.co.il, @haaretz/htz-components, @haaretz/htz-react-server

**Table of Contents**

- [Description](#description)
- [Checklist](#checklist)

**Related issue(s):** #0

## Description

**Notes:**

## Checklist

Design:
  * [ ] Approved by designer

Usability:
  * [ ] Developer
  * [ ] Project manager

Tests
  
  * [ ] UI snapshot testing 
  * [ ] Unit testing
  * [ ] Integration testing
  * [ ] E2E testing


eliagrady:
We might decide to move HTTP logs later on, but for now it's fine IMHO

packages/libs/htz-react-server/src/index.js
```diff
@@ -16,6 +16,7 @@ import config from 'config';
 // modules that Next.js imports while routing and rendering pages.
 import 'isomorphic-fetch';
 import morgan from 'morgan';
+import morganJson from 'morgan-json';
```
TxHawks: Didn't we say that this should really happen in Varnish, and that we should only use morgan in dev?

eliagrady: we did, and therefore we shan't

PR #828, opened 2018-08-15
feat(comments): limit new comment length to 600 chars and double comm…

kedarguy:
…ent body rows before cutting of


packages/components/htz-theme/src/consts/i18n.js
```diff
@@ -95,6 +95,7 @@ export const commentFormI18n = Object.freeze({
   errorNotes: Object.freeze({
     nameErrorNoteTxt: 'חובה להזין שם',
     commentErrorNoteTxt: 'נא להזין את תוכן התגובה',
+    commentErrorToLongNoteTxt: 'אין להזין יותר מ-600 תווים בתוכן התגובה',
```
TxHawks: should be `commentErrorTo0LongNoteTxt` (as in another `o` before the `L` in `Long`

PR #884, opened 2018-09-03
Feature/new sso service

kedarguy:


kedarguy:
We need to make sure this was tested properly

PR #960, opened 2018-10-08
Feature/937 prevent newsletter registration conflict

oren-l:
**Table of Contents**

- [Description](#description)
- [Checklist](#checklist)

**Related issue(s):** #0

## Description

**Notes:**

## Checklist

Design:
  * [ ] Approved by designer

Usability:
  * [ ] Developer
  * [ ] Project manager

Tests
  
  * [ ] UI snapshot testing 
  * [ ] Unit testing
  * [ ] Integration testing
  * [ ] E2E testing


packages/components/htz-components/package.json
```diff
@@ -97,5 +98,8 @@
     "react-fela": "^7.2.0",
     "rimraf": "^2.6.2",
     "svgr": "^1.1.0"
+  },
+  "jest": {
+    "setupTestFrameworkScriptFile": "jest-mock-console/dist/setupTestFramework.js"
```
TxHawks: This means that it is run before _each_ test. Is that really needed? I rather we just import it in tests where is't needed

packages/components/htz-components/src/components/MobileQuickRegistration/MobileQuickRegistration.js
```diff
@@ -114,47 +116,60 @@ class MobileQuickRegistration extends Component {
...
+          {
+            shouldRenderNewsletter => {
+              if (shouldRenderNewsletter) {
+                // prevents this component from being rendered when Newsletter is rendered
+                return null;
+              }
+              return (
+                <FelaTheme
+                  render={theme => (
+                    <Media
```
TxHawks: There is no need for both `<Media />` and `<NoSSR />`. Unless it has a `matchOnServer` attribute, `<Media />` is only rendered in the browser.

packages/components/htz-components/package.json
```diff
@@ -86,8 +86,9 @@
     "chalk": "^2.3.0",
     "enzyme-to-json": "^3.3.4",
     "fela": "^6.1.7",
-    "graphql-tag": "^2.9.2",
     "fela-monolithic": "^5.0.24",
+    "graphql-tag": "^2.9.2",
+    "jest-mock-console": "^0.4.0",
```
TxHawks: This is great! I was looking for something like that for ages. But it should be part of `@haaretz/htz-react-base`, so that all packages can use it out of the box

packages/components/htz-components/src/components/MobileQuickRegistration/MobileQuickRegistration.js
oren-l: The `<NoSSR />` is there to prevent from `<ShouldRenderNewsletterProvider />` from rendering in SSR which would result in a false warning of 'no localStorage of undefined' (as the `<ReadingHistoryProvider />` uses the localStorage)

packages/components/htz-components/package.json
oren-l: Thanks!
I'll move it to htz-react-base

packages/components/htz-components/package.json
oren-l: @TxHawks as you told me later, I understand that this part stays (moves also to htz-react-base)

TxHawks: Since it is handled by `htz-react-base`, this can be removed from here. Same goes for the `jest-mock-console` dependency. But I don't see any changes to `htz-react-base`

packages/components/htz-components/src/components/MobileQuickRegistration/MobileQuickRegistration.js
TxHawks: There should be no problem to remove `<NoSSR />`. Just put `<ShouldRenderNewsletterProvider />` as `<RegistrationWrapper />`'s parent inside the `render` function in `<Media />`

PR #995, opened 2018-10-15
Tool tip

yehudagoldner:
**Table of Contents**

- [Description](#description)
- [Checklist](#checklist)

**Related issue(s):** #0

## Description

**Notes:**

## Checklist

Design:
  * [ ] Approved by designer

Usability:
  * [ ] Developer
  * [ ] Project manager

Tests
  
  * [ ] UI snapshot testing 
  * [ ] Unit testing
  * [ ] Integration testing
  * [ ] E2E testing


packages/components/htz-components/src/components/ActionButtons/actionList.js
```diff
@@ -580,38 +581,58 @@ const Whatsapp: StatelessFunctionalComponent<WhatsappButtonProps> = ({
...
-          biAction({
-            actionCode: biActionMapper.get('whatsApp_share'),
-            additionalInfo: {
-              platform,
-            },
-          });
-          return false;
-        }}
+      <Tooltip
+        text="שתפו וחבריכם יקראו את הכתבה ללא תשלום"
```
TxHawks: This text should not be hard-coded.

It should either come from Polopoly or the theme's i18n.

packages/components/htz-components/src/components/Tooltip/Tooltip.js
```diff
@@ -0,0 +1,279 @@
...
+import IconClose from '../Icon/icons/IconClose';
+import { stylesPropType, } from '../../propTypes/stylesPropType';
+import { attrsPropType, } from '../../propTypes/attrsPropType';
+
+const close = (theme, side) => ({
+  height: '100%',
+  width: '5.8rem',
+  position: 'absolute',
+  left: side === 'RIGHT' ? 'auto' : '0',
+  right: side === 'RIGHT' ? '0' : 'auto',
```
TxHawks: Use `inlineStart` and `inlineEnd` here so that it's layout direction independent

packages/components/htz-components/src/components/Tooltip/Tooltip.js
```diff
@@ -0,0 +1,279 @@
...
+    attrs: attrsPropType,
+  };
+
+  static defaultProps = {
+    openSide: 'auto',
+    closingCount: 2,
+    miscStyles: null,
+    hide: false,
+    storageParam: null,
+    tooltipMiscStyles: () => [ {}, ],
```
TxHawks: As mentioned before, I think it is better if this isn't a function, but even if it is, better set it to null by default that running a noop

packages/components/htz-components/src/components/Tooltip/Tooltip.js
```diff
@@ -0,0 +1,279 @@
...
+class Tooltip extends Component {
+  static propTypes = {
+    /** The text of the tooltip */
+    text: PropTypes.string.isRequired,
+    /** The local storage parameter */
+    storageParam: PropTypes.string,
+    /** Amount of times to show the tooltip after closing */
+    closingCount: PropTypes.number,
+    /** horizentol openning side (auto | TOP | BOTTOM | LEFT | RIGHT) */
+    openSide: PropTypes.string,
```
TxHawks: Why in CAPS? We don't generally do this with `string` props.
Also, `TOP` and `BOTTOM` are not horizontal, and it's better to use `inlineStart` and `inlineEnd` rather than `left` and `right`, so that it is layout-direction independent.

packages/components/htz-components/src/components/Tooltip/Tooltip.js
```diff
@@ -0,0 +1,279 @@
...
+    offsetX,
+    tooltipMiscStyles,
+    openSide,
+  }) => ({
+    ...tooltipPosition,
+    position: 'absolute',
+    zIndex: '1',
+    display: hide ? 'none' : 'inline-block',
+    backgroundColor: theme.color('primary', 'base'),
+    color: '#FFF',
```
TxHawks: All colors should come from the theme

packages/components/htz-components/src/components/Tooltip/Tooltip.js
```diff
@@ -0,0 +1,279 @@
...
+
+  hideTooltip = () => {
+    this.setState({ hide: true, });
+  };
+
+  calcPadding = openSide => {
+    const closeButtonSide = this.props.openOnMouseOver ? '3rem' : '8rem';
+    return {
+      paddingLeft: openSide === 'RIGHT' ? '3rem' : closeButtonSide,
+      paddingRight: openSide === 'RIGHT' ? closeButtonSide : '3rem',
```
TxHawks: `paddingInlineStart` and `paddingInlineEnd`

packages/components/htz-components/src/components/Tooltip/Tooltip.md
```diff
@@ -0,0 +1,41 @@
...
+<div style={{ margin: '200px 0 0 300px' }}>
+  <Tooltip text="What is up?">
+    <p style={{ margin: '0px', display: 'inline-block' }}>goodby</p>
+  </Tooltip>
+</div>
+```
+
+## Openning side
+
+Forcing the tooltip openning side. Sccepts : TOP | BOTTOM | LEFT | RIHGT | AUTO (default)
```
TxHawks: `Accepts`. Also, notes about explicit direction vs `inlineStart`|`end` apply here as well. of course

packages/components/htz-components/src/components/Tooltip/Tooltip.js
```diff
@@ -0,0 +1,171 @@
...
+/* global document */
+import React, { Component, } from 'react';
+import { FelaComponent, } from 'react-fela';
+import PropTypes from 'prop-types';
+import IconClose from '../Icon/icons/IconClose';
+
+const tooltip = ({ theme, tooltipPosition, arrowStyle, hide, offset, }) => ({
+  ...tooltipPosition,
+  position: 'absolute',
+  zIndex: '100',
```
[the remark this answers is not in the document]
yehudagoldner: Until now the z-index haven't been managed from one place. So doing it only here would be meaningless.
If we decide to do such thing it should be across the project, is should be separated as different brunch and different task.

packages/components/htz-components/src/components/Tooltip/Tooltip.js
yehudagoldner: If I don't use the style as a function then who uses this component need to get the theme object somewhere else. It's not comfortable

Issue #1011, opened 2018-10-17
Image - "credit" fix


adibi83:
## Issue Type

**Update to existing feature**

## Description
In image credit - "צילום" instead of "קרדיט"

PR #1013, opened 2018-10-17
fix(image_caption_prefix): changing caption prefix only for images

TawKhoury:

- [Description](#description)

**Related issue(s):** #1011 

## Description
Changing credit prefix.


packages/components/htz-components/src/components/ArticleBodyImage/ArticleBodyImage.js
```diff
@@ -282,7 +282,7 @@ const UnwrappedImage = ({
       <figure className={className} onClick={!isFullScreen ? toggleFullScreen : null}>
         <ImageElement forceAspect={forceAspect} {...image} isFullScreen={isFullScreen} />
         {showCaption && !isFullScreen ? (
-          <Caption caption={title} credit={credit} prefix="צילום" />
+          <Caption caption={title} credit={credit} creditprefix="צילום" />
```
TxHawks: This should really come from the theme. It's not going to be צילום in the English edition

TawKhoury: Done! Thanks

PR #1054, opened 2018-10-28
Fix/purchase page pay with existing card

kedarguy:


packages/apps/purchase-page/components/OfferPage/Stages/ExistingCreditCardStage.js
```diff
@@ -122,7 +122,7 @@ class Wrapper extends Component {
       description: paymentData.description,
       failure: false,
       // does not matter what we send so hardcoded
-      connectionType: '764',
+      connectionType: 764,
```
TxHawks: This is a bit confusing, because, first, it seem to does matter what send, since we had to convert the `string` to a `number`, and more importantly, second, if it doesn't matter and doesn't ever change, why not just do it in the resolver?

Issue #1081, opened 2018-11-07
Navigation menu - pusr


adibi83:
## Issue Type

**Update to existing feature**

## Description
Navigation menu:
If the user is a pusr, don't show the "רכשו מינוי" button

PR #1085, opened 2018-11-07
feat(nav_menu): omit purchase now option in nav menu when htzpusr exists

TawKhoury:

**Table of Contents**

- [Description](#description)
- [Checklist](#checklist)

**Related issue(s):** #1081 

## Description

**Notes:**

## Checklist

Design:
  * [ ] Approved by designer

Usability:
  * [ ] Developer
  * [ ] Project manager

Tests
  
  * [ ] UI snapshot testing 
  * [ ] Unit testing
  * [ ] Integration testing
  * [ ] E2E testing


packages/components/htz-components/src/components/NavigationMenu/NavigationMenu.js
```diff
@@ -120,7 +117,17 @@ class NavigationMenu extends React.Component {
     }).isRequired,
   };
 
-  state = { isHovered: false, };
+  state = { isHovered: false, isUser: null, };
+
+  componentDidUpdate() {
+    if (window && typeof window !== 'undefined') {
```
TxHawks: If anything, this check should be reversed, but you don't need both.
```suggestion
    if (typeof window !== 'undefined') {
```

But do we even need it at all? `componentDidUpdate` isn't run on the server, so `window` is guaranteed to exist.

packages/components/htz-components/src/components/NavigationMenu/NavigationMenu.js
```diff
@@ -120,7 +117,17 @@ class NavigationMenu extends React.Component {
...
+      const isUser = typeof CookieUtils.getCookie('HtzPusr') === 'string';
```
TxHawks: Where is `CookieUtils` coming from?

Also, it seems like you are checking if it's a _paying_ user, not just any user. The name is confusing.

packages/components/htz-components/src/components/NavigationMenu/NavigationMenu.js
```diff
@@ -120,7 +117,17 @@ class NavigationMenu extends React.Component {
...
+      if (this.state.isUser !== isUser) {
+        /* eslint-disable react/no-did-update-set-state */
+        this.setState({ isUser, });
```
TxHawks: Why do this `setState` in `componentDidUpdate`? It will result in a re-render

packages/components/htz-components/src/components/NavigationMenu/NavigationMenu.js
```diff
@@ -1,4 +1,4 @@
-/* global document */
+/* global document, window */
```
TxHawks: The `document` global is never used

TawKhoury: Used in line 142

PR #1133, opened 2018-11-18
feat(watch): add watch command to packages for better dx

yehudagoldner:
affects: @haaretz/haaretz.co.il, @haaretz/htz-components, @haaretz/htz-theme, @haaretz/tm-theme,
@haaretz/app-utils, @haaretz/d3-server, @haaretz/dfp, @haaretz/fela-utils, @haaretz/graphql-server,
@haaretz/htz-css-tools, @haaretz/htz-react-base, @haaretz/htz-react-server, @haaretz/htz-user-utils

**Table of Contents**

- [Description](#description)
- [Checklist](#checklist)

**Related issue(s):** #0

## Description

**Notes:**

## Checklist

Design:
  * [ ] Approved by designer

Usability:
  * [ ] Developer
  * [ ] Project manager

Tests
  
  * [ ] UI snapshot testing 
  * [ ] Unit testing
  * [ ] Integration testing
  * [ ] E2E testing


package.json
```diff
@@ -28,6 +28,8 @@
     "test": "lerna exec --no-bail -- yarn run test",
     "test:deploy": "lerna exec -- yarn run test --forceExit",
     "update": "yarn install && lerna run --no-bail --stream prepare",
+    "htz:watch": "yarn workspace @haaretz/haaretz.co.il watch",
+    "h1": "yarn workspace @haaretz/haaretz.co.il dev",
```
TxHawks: These belong in the individual packages, not here

package.json
```diff
@@ -452,7 +454,8 @@
     "concurrent": false
   },
   "devDependencies": {
-    "@commitlint/cli": "^4.1.0",
+    "@babel/preset-flow": "^7.0.0",
+    "@commitlint/cli": "^5.1.0",
```
TxHawks: Why the update to `@commitlint/cli` in this pull request?

packages/apps/haaretz.co.il/package.json
```diff
@@ -21,7 +21,10 @@
     "styleguide:build": "htz-scripts styleguide:build",
     "test": "htz-scripts test",
     "update-scripts": "htz-scripts update-scripts",
-    "dev": "cross-env concurrently \"cross-env NODE_ENV=development graphql-server\" \"cross-env NODE_ENV=development htz-react-server htz | bunyan -L\"",
+    "watch": "cross-env concurrently \"yarn workspace @haaretz/htz-components watch\" \"yarn workspace @haaretz/htz-theme watch\" \"yarn workspace @haaretz/app-utils watch\" \"yarn workspace @haaretz/htz-react-server watch\" \"yarn workspace @haaretz/graphql-server watch\"",
```
TxHawks: I'd call this one `watchExternals` or something like that. Also, no need to watch `htz-react-server` and `graphql-server`, they cannot be hot-reloaded.

packages/apps/haaretz.co.il/package.json
```diff
@@ -21,7 +21,10 @@
...
+    "dev": "cross-env concurrently \"cross-env NODE_ENV=development graphql-server\" \" cross-env NODE_ENV=development htz-react-server htz | bunyan -L\"",
+    "dev:nodemoon": "cross-env concurrently \"cross-env NODE_ENV=development graphql-server\" \" cross-env NODE_ENV=development nodemon ../../libs/htz-react-server/dist/lib/index.js htz | bunyan -L\"",
```
TxHawks: It's `nodemon`, you have an extra `o` there...

But I think the naming isn't expressive enough. It exposes an implementation detail (the task uses `nodemon` internally) instead of explaining to the user what it does.

Also, I don't see the point in having `dev`, `dev:nodemon` and `dev:watch`. As far as I can tell, they all have the same use case, which is firing up a development server, no? I feel like this is redundent API surface

packages/libs/htz-react-base/scripts/build/lib.js
```diff
@@ -8,7 +8,10 @@ checkDir('dist').then(hasDist => {
   if (!hasDist) {
     fs.mkdirSync('dist');
   }
-  if (process.argv.length < 3) {
+  if (process.argv.includes('watch')) {
```
TxHawks: I'd reverse the logic - if `argv.length < 3` it inherently means there is no watch argument.

packages/libs/htz-react-base/scripts/build/lib.js
```diff
@@ -8,7 +8,10 @@ checkDir('dist').then(hasDist => {
...
+    process.argv = [ process.argv[0], process.argv[1], '--watch', 'src', '--out-dir', 'dist/lib', ];
```
TxHawks: I'd do this a bit differently to preserve flexibility:
 
```suggestion
    const watchIndex = process.argv.indexOf('watch');
    const const [ executer, command, ...args ] = process.argv;
    if (process.argv.length > 4) {  
      args.splice(watchindex, 1);
      process.argv = [ executer, command, '--watch', ...args ];
    }
    else {
       process.argv = [ executer, command, '--source-maps', '--watch', 'src', '--out-dir', 'dist/lib', ];
    }
```

But please check this snippet before applying it, because I haven't tested it

packages/apps/haaretz.co.il/package.json
```diff
@@ -21,7 +21,9 @@
     "styleguide:build": "htz-scripts styleguide:build",
     "test": "htz-scripts test",
     "update-scripts": "htz-scripts update-scripts",
-    "dev": "cross-env concurrently \"cross-env NODE_ENV=development graphql-server\" \"cross-env NODE_ENV=development htz-react-server htz | bunyan -L\"",
+    "watchExternals": "cross-env concurrently \"yarn workspace @haaretz/htz-components watch\" \"yarn workspace @haaretz/htz-theme watch\" \"yarn workspace @haaretz/app-utils watch\" \"yarn workspace @haaretz/htz-react-server watch\" \"yarn workspace @haaretz/graphql-server watch\"",
+    "dev": "cross-env concurrently \"cross-env NODE_ENV=development graphql-server\" \" cross-env NODE_ENV=development htz-react-server htz | bunyan -L\"",
+    "dev:watch": "cross-env concurrently \"yarn workspace @haaretz/haaretz.co.il dev\" \"yarn workspace @haaretz/haaretz.co.il watch\"",
```
TxHawks: If this is intended to be the main script for dev, it should just be `dev`, not `dev:watch`. To the best of my understanding, the script currently named `dev` is only there so that it can by internally used by this script, so maybe we can rename it to `dev:internal` or something.

package.json
```diff
@@ -470,10 +471,10 @@
     "lint-staged": "^7.2.0",
     "lodash.template": "^4.4.0",
     "node-dir": "^0.1.17",
+    "nodemon": "^1.18.6",
```
TxHawks: I think we don't need this dependency any more, right?

packages/apps/haaretz.co.il/package.json
yehudagoldner: i just removed the dev:watch it's not really important

packages/apps/haaretz.co.il/package.json
```diff
@@ -21,7 +21,9 @@
     "styleguide:build": "htz-scripts styleguide:build",
     "test": "htz-scripts test",
     "update-scripts": "htz-scripts update-scripts",
-    "dev": "cross-env concurrently \"cross-env NODE_ENV=development graphql-server\" \"cross-env NODE_ENV=development htz-react-server htz | bunyan -L\"",
+    "watchExternals": "cross-env concurrently \"yarn workspace @haaretz/htz-components watch\" \"yarn workspace @haaretz/htz-theme watch\" \"yarn workspace @haaretz/app-utils watch\" \"yarn workspace @haaretz/htz-react-server watch\" \"yarn workspace @haaretz/graphql-server watch\"",
```
TxHawks: Like I said in the previous comment, there is no need to watch `htz-react-server` and `graphql-server`. You'd need to restart your dev server anyway if they change.

packages/apps/haaretz.co.il/package.json
TxHawks: But it's still here...

But I actually think it's good to keep it, that way you don't have to manually execute both `dev` and `watchExternal`. Or am I missing something?

packages/apps/haaretz.co.il/package.json
yehudagoldner: but if i have this watch and i change something there i can only restart the app without building the lib

packages/libs/htz-react-base/scripts/build/lib.js
TxHawks: The way you did it doesn't let us override default arguments like `--source-maps` when passing `watch`.

Try something like this:

```js
const { argv, } = argv;
// user did not pass any arguments
if (argv.length < 3) {
  argv.push('src', '--source-maps', '--out-dir', 'dist/lib');
}
// User passed arguments, and we need to check if `watch` is one of them
else {
  const watchIndex = argv.indexOf('watch');
  const hasWatch = watchIndex !== -1;

  if (hasWatch) {
    argv[watchindex] = '--watch';

    // User only passed the `watch` argument, add defaults
    if (argv.length < 4) {
      argv.push(
        '--source-maps',
        'src',
        '--out-dir',
        'dist/lib'
     ); 
    }
  }
}
```

But again, I didn't test this, so please validate that it works as expected

Issue #1169, opened 2018-11-26
Review article comments bug


adibi83:
## Issue Type

**Bug fix**

## Description
No comments in review article (react-prod):
https://react-prod.haaretz.co.il/REVIEW-1.6676696

PR #1174, opened 2018-11-26
Fix/comments not loading

kedarguy:


kedarguy:
closes #1169

This bug was not just for review articles, happened in all other articles except standard

PR #1376, opened 2019-01-09
feat(lists deafult image): add a default image component and use in a…

kedarguy:
…ll lists when there is no image


kedarguy:
@TxHawks I put 'neutral', '-4' as the placeholder color, need to decide which color to put there

packages/components/htz-components/src/components/Image/DefaultImage.js
```diff
@@ -0,0 +1,42 @@
...
+  const aspectRatio = width && height ? { width, height, } : aspectRatios[aspect];
+
+  const paddingPercentage = (aspectRatio.height / aspectRatio.width) * 100;
+  return {
+    paddingTop: `${paddingPercentage}%`,
+  };
+}
+
+export default function DefaultImage({ aspect, width, height, }: Props) {
+  if (!aspect && !(!width || !height)) {
```
TxHawks: This should be:
```suggestion
  if (!aspect && !(width && height)) {
```

Which reads clearer to me, because it says, if `no aspect` and `no (width AND height)`, which is what we are checking. 

It is the same, but seems less convoluted to me, and the proliferation of bangs and the `or` operator obfuscate intent

packages/components/htz-components/src/components/Image/DefaultImage.js
```diff
@@ -0,0 +1,42 @@
...
+    console.warn(
```
TxHawks: I think this should actually throw. It is a hard failure of the component

packages/components/htz-components/src/components/Image/Image.js
```diff
@@ -181,7 +173,9 @@ Please use the "<Picture />" component`
 
   render() {
     if (this.props.data === null) {
-      return null;
+      const { transforms, } = this.props.imgOptions;
+      const aspect = Array.isArray(transforms) ? transforms[0].aspect : transforms.aspect;
+      return <DefaultImage aspect={aspect} />;
```
TxHawks: We need to also pass `width` and `height` here

packages/components/htz-components/src/components/Image/DefaultImage.js
```diff
@@ -0,0 +1,29 @@
...
+import { parseComponentProp, } from '@haaretz/htz-css-tools';
+import { aspectRatios, } from './Image';
+
+type Props = {
+  aspect: string | ComponentPropResponsiveObject<string>[],
+};
+
+const style = ({ theme, aspect, }) => ({
+  // todo: ask designer to approve/change color
+  backgroundColor: theme.color('neutral', '-4'),
```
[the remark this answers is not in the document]
kedarguy: done

packages/components/htz-components/src/components/Image/DefaultImage.js
kedarguy: done

packages/components/htz-components/src/components/Image/DefaultImage.js
kedarguy: done

packages/components/htz-components/src/components/Image/Image.js
kedarguy: done

PR #1538, opened 2019-01-29
Feature/table score

EnoshTs:
**Table of Contents**

- [Description](#description)
- [Checklist](#checklist)

**Related issue(s):** #0

## Description

**Notes:**

## Checklist

Design:
  * [ ] Approved by designer

Usability:
  * [ ] Developer
  * [ ] Project manager

Tests
  
  * [ ] UI snapshot testing 
  * [ ] Unit testing
  * [ ] Integration testing
  * [ ] E2E testing

**Sport's Results Table Score Component**

**TableScore** is an abstract component that creates a custom table score

_First argument is **tableType** that filters results by sport's type_

**And should be only:**

1.  **nba**
2. **soccer-leagues**
3. **soccer-champions**

_Second argument depend on **tableType**_

**nba**  :
**coastType**  should be only - east/west:
Defines which coast will display first ( switchable )

**soccer-leagues** :
**league** should be only 2125, 2021, 2019, 2015, 2014, 2002
Defines which league will display ( permanent )

**soccer-champions**  :
**number** should be between 1 - 8
Defines which champion group will display first ( switchable )

_Third argument ( optional ) **isOpen**_

Only available for **nba** and **soccer-leagues** !
When **isOpen** present - it displays full data table score
Otherwise displays only half of the data
**isOpen** is switchable by a toggle button

Examples:

**nba**
`<TableScore tableType="nba" coastType="west" isOpen/>` 
`<TableScore tableType="nba" coastType="east" />` 

**soccer-leagues**
`<TableScore tableType="soccer-leagues" league="2125" isOpen/>` 
`<TableScore tableType="soccer-leagues" league="2002" />` 

**soccer-champions**
`<TableScore tableType="soccer-champions" number="5" />`


TxHawks:
A few notes on accessibility:
#### Table Accesibility
1. switching between the East Coast and West Coast tables should be done using our `<Tabs />` component.
2. A table's first child should always be a [`<caption /`>](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/caption) element containing the table's title. If it's not desired for it to be visible for sighted users, you can wrap its content with our `<AriaDescription />` component.
3. column and row headers should be explicitly scoped using `<th>`'s `scope` attribute. See [here for details](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/table#Scoping_rows_and_columns)

#### Load More
1. The table should have a `tabIndex` attribute with a value of `-1` so that it is outside of the tabable focus flow, but can be focused programatically.
2. The `load more` buttons should
    * have a `type` attribute of `button`
    * Have an `aria-expanded` attribute, that is set to `false` as long as the table is collapsed, and changed to `true` when it is expanded.
    * Have an `aria-controls` attribute, that points to the table by its `ID`.
3. On expanding the table,
    * the `aria-expanded` attribute should be changed to 'true'
    * focus should be moved to the table
4. On collapsing the table, focus should not be moved, but the `aria-expanded` attribute should be changed to `false`.

PR #1673, opened 2019-02-20
feat(survey): wrap the date of the legend with b tag

yehudagoldner:
**Related issue(s):** #0

## Description

**Notes:**

## Checklist

  * [ ] Approved by designer

Usability:
  * [ ] Developer
  * [ ] Project manager

Tests
  
  * [ ] UI snapshot testing 
  * [ ] Unit testing
  * [ ] Integration testing
  * [ ] E2E testing

### DIFF
--- a/packages/components/htz-components/src/components/Survey/Survey.js
+++ b/packages/components/htz-components/src/components/Survey/Survey.js
@@ -3,7 +3,6 @@ import React from 'react';
 import { FelaComponent, FelaTheme, } from 'react-fela';
 import orderBy from 'lodash/orderBy';
 import Observer from 'react-intersection-observer';
-
 import style from './style';
 import Button from '../Button/Button';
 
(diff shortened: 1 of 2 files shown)


packages/components/htz-components/src/components/Survey/Survey.js
```diff
@@ -107,18 +107,25 @@ export default class Survey extends React.Component {
...
+                        return (
+                          <FelaComponent style={style.legend} key={this.colors[i]}>
+                            <FelaComponent
+                              color={this.colors[i]}
+                              rule={style.legendcolor}
+                              render="span"
+                            />
+                            <FelaComponent style={style.legendLabel} render="span">
+                              <span>{legend[0]}</span>
+                              <b>{legend[1]}</b>
```
TxHawks: Why `<b>`? For styling the text bold? It's not a good idea to do that with an HTML element, as it is prone to change with a reset. The tag also has a different semantic meaning to it. 
The `<b>` element in its old presentational bold meaning has been deprecated, and now has semantic meaning of bring-attention-to-element, which is probably not what you want here. You probably want <strong> here, and with `font-weight: 700`.

Issue #1717, opened 2019-03-04
Split field text into 2 lines with |


innbal:
Should be applied in product and price pages

chenaviv:
Lines should break with a line break, instead of the '|' character.
Pull request is pending for a fix for this on the front-end.

PR #1719, opened 2019-03-06
fix(offer text): support multiple lines of offer text

chenaviv:
affects: @haaretz/purchase-page

offer.text is an array of strings, but seems to have been treated as a single string.
This caused multiple strings to not be iterated over when mapped to jsx, but rather be treated as a single item.

**Related issue(s):** #1717

## Description

Amended - removed fixed height definitions and slightly changed the padding for the smaller box in large and xl views.

**Notes:**

Note that over two lines of text will currently result in an overflow of content outside of the offer box in viewports 600px wide and over, as the height of the offer box is not dynamic in those views.

## Checklist

  * [ ] Approved by designer

Usability:
  * [ ] Developer
  * [ ] Project manager

Tests
  
  * [ ] UI snapshot testing 
  * [ ] Unit testing
  * [ ] Integration testing
  * [ ] E2E testing


packages/apps/purchase-page/components/OfferPage/Stages/ChooseProductStageElements/Offer.js
```diff
@@ -274,7 +276,7 @@ const DesktopOffer = ({
                 </StyledOfferTitle>
                 <StyledPrice isRecommended={offer.isRecommended}>{offer.price}</StyledPrice>
                 {(offer.text || offer.originalPrice)
-                  && [ offer.originalPrice, ...(offer.text ? [ offer.text, ] : []), ].map(
+                  && [ offer.originalPrice, ...(offer.text || []), ].map(
                     (text, textIndex) => (
                       <StyledOfferText
                         key={Math.random()}
```
TxHawks: This isn't your code, but if we're already touching this, let's fix this.

`key` should always be both unique AND stable, so `Math.random()` doesn't work here. If I'm not mistaken, the value of `text` should be okay here. it is definitely stable, and should be unique

chenaviv: Changed the key to be `text`

TxHawks: thanks

PR #1727, opened 2019-03-10
feat(surveys): float support

yehudagoldner:

**Related issue(s):** #0

## Description

**Notes:**

## Checklist

  * [ ] Approved by designer

Usability:
  * [ ] Developer
  * [ ] Project manager

Tests
  
  * [ ] UI snapshot testing 
  * [ ] Unit testing
  * [ ] Integration testing
  * [ ] E2E testing


packages/components/htz-components/src/components/Survey/Survey.js
```diff
@@ -46,7 +46,7 @@ export default class Survey extends React.Component {
 
       // eslint-disable-next-line array-callback-return
       this.xls.surveys.map(item => {
-        unsortedItems.map(v => this.partiesValues[v].push(parseInt(item[v], 10)));
+        unsortedItems.map(v => this.partiesValues[v].push(+(+item[v]).toFixed(1)));
```
TxHawks: Two things about these lines:
The use of `map` to perform side effects is confusing, since map returns a new array, implying the code is interested in the return value. That's also the reason for the eslint rule you disabled. Any reason not to use `forEach`? Seems more suitable for something like this.

I don't really see the point in coercing `number` to `string` and back again. I think it'd be simpler to do:
```js
Math.round(item[v] * 10) / 10
```
or you could even create a generic function:
```js
function roundToDecimalPrecision(num, precision = 0) {
  const multiplier = Math.pow(10, precision)
  return Math.round(num * multiplier) / multiplier;
}
```

and then just push `roundToDecimalPrecision(item[v], 1)`

PR #1746, opened 2019-03-21
feat(components): add iframe element

EranM6:
affects: @haaretz/haaretz.co.il, @haaretz/htz-components, @haaretz/app-utils

**Related issue(s):** #0

## Description

**Notes:**

## Checklist

  * [ ] Approved by designer

Usability:
  * [ ] Developer
  * [ ] Project manager

Tests
  
  * [ ] UI snapshot testing 
  * [ ] Unit testing
  * [ ] Integration testing
  * [ ] E2E testing


packages/components/htz-components/src/components/Iframe/Iframe.js
```diff
@@ -0,0 +1,53 @@
...
+}
+
+class Iframe extends React.Component<IframeType, State> {
+  state = {
+    shouldRender: false,
+    height: `${this.props.height}px`,
+  };
+
+  componentDidMount() {
+    window.addEventListener('message', this.receiveMessage, false);
```
TxHawks: You forgot to remove the event listener on unmount. Let's just convert this class to a function component with hooks, it handles this more cleanly

packages/components/htz-components/src/components/Iframe/Iframe.js
```diff
@@ -0,0 +1,53 @@
...
+  iframeRef: ElementRef<'iframe'> | null;
+
+  render(): Node {
+    const { src, contentName, } = this.props;
+    const { shouldRender, height, } = this.state;
+
+    return shouldRender
+      ? (
+        <iframe
+          ref={iframeRef => { this.iframeRef = iframeRef; }}
```
TxHawks: This is the old api. Better use `createRef()` or `useRef()`. But actually, why do you need a ref here?

packages/components/htz-components/src/components/Iframe/Iframe.js
```diff
@@ -0,0 +1,51 @@
...
+type State = {
+  shouldRender: boolean,
+  height: string,
+}
+
+function Iframe({ src, contentName, elementId, ...props }: IframeType): Node {
+  const [ state, setState, ] = useState({
+    shouldRender: false,
+    height: `${props.height}px`,
+  });
```
TxHawks: In this case, I'd split this to two separate `useState` calls, because they aren't really connected to each other. A good sign that this is a good idea is that whenever you call `setState` the only reason you need its callback signature is to merge the touched value with the untouched value.
```js
const [shouldRender, setShouldRender, ] = useState(false);
const [ height, setHeight, ] = useState(`${props.height}px`);
```

would work better here. It will then allow you to do
```js
setHeight(event.data.height)
```

and
```js
setShouldRender(true)
```

packages/components/htz-components/src/components/Iframe/Iframe.js
```diff
@@ -0,0 +1,51 @@
...
+        // $FlowFixMe
+        height: event.data.height,
+      }));
+    }
+  };
+
+  useEffect(() => {
+    window.addEventListener('message', receiveMessage, false);
+    setState((prevState: State) => ({ ...prevState, shouldRender: true, }));
+  });
```
TxHawks: You need to return a clean up function that will run when it's time to remove the event listener, also you need to specify an empty dependencies array, which ensures this only runs once, on mount:
```suggestion
    return () => window.removeEventListener('message', receiveMessage)
  }, []);
```

packages/apps/haaretz.co.il/utils/getHomePageElements.js
```diff
@@ -73,6 +74,7 @@ const elements = new Map([
   [ 'Morbo', Morbo, ],
   [ 'Panucci', Panucci, ],
   [ 'Pazuzu', Pazuzu, ],
+  [ 'Roberto', Roberto, ],
```
TxHawks: Roberto is really very rarely used, so lets only dynamically import it when needed.

My suggestion is that if the element in the map is a component, use the component, if it's a `string`, use it to dynamically import from `htz-components`. So here change to:
```suggestion
  [ 'Roberto', 'Roberto', ],
```
and then change `getHomepageElements` to:
```diff
export default function getHomepageElements(inputTemplate) {
+  const element = elements.get(inputTemplate) || DefaultComponent;
+  const Component = typeof element === 'string'
+    ? dynamic(() => import('@haaretz/htz-components').then(module => module[element]
+    : element;
+  return Component;
-  return elements.get(inputTemplate) || DefaultComponent;	  return elements.get(inputTemplate) || DefaultComponent;
}	}
```

I'm open to other suggestions on how to do this though

packages/apps/haaretz.co.il/utils/getHomePageElements.js
```diff
@@ -31,6 +31,7 @@ import {
   Morbo,
   Panucci,
   Pazuzu,
+  Roberto,
```
TxHawks: This is only very rarely used. Let's only conditionally import this (see comment below)
```suggestion
```

PR #1747, opened 2019-03-21
Feature/permutive article page

TawKhoury:
**Related issue(s):** #0

## Description

**Notes:**

## Checklist

  * [ ] Approved by designer

Usability:
  * [ ] Developer
  * [x] Project manager

Tests
  
  * [ ] UI snapshot testing 
  * [ ] Unit testing
  * [ ] Integration testing
  * [ ] E2E testing


packages/apps/haaretz.co.il/layouts/ArticleLayout.js
```diff
@@ -150,11 +152,15 @@ class ArticleLayout extends React.Component {
               ? standardArticleElement.isPremiumContent
               : null;
 
-            const authorsArr = slots.article.find(item => item.authors);
+            const slotsArticleDataArr = slots.article.find(item => item.authors);
```
TxHawks: If you do something like this instead:
```js
            const {
              authors = [],
              tags = [],
              // ...
            } = slots.article.find(item => !!item.__typename === 'ArticleData') || [];
```

You don't need to check every time, you can simply map to a new array

PR #1780, opened 2019-04-03
feat(abuse-service): check abuse status upon access to an article

chenaviv:
affects: @haaretz/haaretz.co.il, @haaretz/htz-components, @haaretz/htz-user-utils

If user is classified as an abuser, log out and redirect to the appropriate login page.

**Related issue(s):** #0

## Description

Uses the previously written 'abuse-service' with a few adjustments.
In user-service.js: when dealing with the plant-cookie/images strategy - replaced Promise.race with a promiseOnce function, because the race would always result in a rejection (one of the images fails to load).

**Notes:**

## Checklist

  * [ ] Approved by designer

Usability:
  * [ ] Developer
  * [ ] Project manager

Tests
  
  * [ ] UI snapshot testing 
  * [ ] Unit testing
  * [ ] Integration testing
  * [ ] E2E testing


TxHawks:
Waiting for status update from @innbal

PR #1837, opened 2019-05-01
Feature/algolia search

avipk:
**Related issue(s):** #0

## Description

**Notes:**

## Checklist

  * [ ] Approved by designer

Usability:
  * [ ] Developer
  * [ ] Project manager

Tests
  
  * [ ] UI snapshot testing 
  * [ ] Unit testing
  * [ ] Integration testing
  * [ ] E2E testing


TxHawks:
A few general things:
* The input should probably be focused by default, so that users can just start typing. Lets add a `focusableEl` prop to the modal component, which will take a ref.
* There should not be an outline on the focused input, the blinking cursor is indication enough
* The search modal redirects to the wrong page when search is performed anywhere but the homepage (redirects relatively)

PR #1841, opened 2019-05-01
fix(init pixel scripts): move init pixel to createdocument

TawKhoury:
affects: @haaretz/haaretz.co.il, @haaretz/purchase-page, @haaretz/htz-components

createapp render scripts twice while initpixel component wrap all scripts with next head component :
based on nextjs issue 5697

**Related issue(s):** #0

## Description

**Notes:**

## Checklist

  * [ ] Approved by designer

Usability:
  * [ ] Developer
  * [ ] Project manager

Tests
  
  * [ ] UI snapshot testing 
  * [ ] Unit testing
  * [ ] Integration testing
  * [ ] E2E testing


packages/components/htz-components/src/createDocument.js
```diff
@@ -155,7 +154,7 @@ const createDocument = ({
 
             {/* ChartBeat scripts should only render on homepage */}
             {path !== '/' ? null : this.ChartbeatConfig()}
-
+            {AdditionalScripts ? <AdditionalScripts host={this.props.host} /> : null}
```
TxHawks: `additionalScripts` should be an array, so that it's possible to pass more than one additional script.

packages/components/htz-components/src/createDocument.js
```diff
@@ -41,6 +39,7 @@ const createDocument = ({
   lang,
   hasToggleableTheme = false,
   fontStacks,
+  AdditionalScripts,
```
TxHawks: Since it's an array, it should be `additionalScripts`, PascalCase indicates it's a component

TawKhoury: Done, thanks.

Issue #1940, opened 2019-06-02
Add Twitter meta 


adibi83:
## Issue Type

**New Feature**

## Description
Add Twitter meta to all articles:
[TWITTER META.txt](https://github.com/Haaretz/htz-frontend/files/3244649/TWITTER.META.txt)

PR #1941, opened 2019-06-02
feat(articles meta): add twitter meta to all article types

kedarguy:

ISSUES CLOSED: #1940


packages/components/htz-components/src/components/ArticleTypes/LiveBlogArticle/LiveBlogArticle.js
```diff
@@ -116,6 +116,11 @@ function LiveBlog({ articleId, slots, path, isBlock, isExploded, }) {
                     <meta property="og:image:width" content="1200" />
                     <meta property="og:image:height" content="630" />
                     <meta property="ob:title" content={obTitle} />
+                    <meta name="twitter:card" content={ogImageUrl} />
+                    <meta name="twitter:site" content="@haaretz" />
```
TxHawks: Doesn't this change for the marker and the english site?

Also, since it all repeats itself in all article types, probably best to extract it to its own component

kedarguy: added TwitterMeta component and based twitter handler on site prop

PR #1951, opened 2019-06-04
feat(magazine article): add photoblog mode and refactor some componen…

kedarguy:
…ts to hooks and flow

affects: @haaretz/haaretz.co.il, @haaretz/htz-components, @haaretz/graphql-server


kedarguy:
need to update papi  to output isPhotoBlog and test that the feature is working with correct data

kedarguy:
@TxHawks  ready for merge

Issue #1886, opened 2019-05-16
Scruffy List 


adibi83:
## Issue Type

**New Feature**

## Description
- List title
-	image
-	Exclusive title
-	Title
-	Author
-	Update date
-	Comments count

Design (see list "מיוחדים"):
XL - https://app.zeplin.io/project/59b4e63a7a92547269089328/screen/5c1f62111c6890243da92cf1

L - 
https://app.zeplin.io/project/59b4e63a7a92547269089328/screen/5c1f62193e193baefe7639f9

M - 
https://app.zeplin.io/project/59b4e63a7a92547269089328/screen/5c1f6214a493040a8b84f944

S - 
https://app.zeplin.io/project/59b4e63a7a92547269089328/screen/5c1f62136309de247970ebf0

PR #1968, opened 2019-06-10
feat(santa list): add santa list

kedarguy:
affects: @haaretz/haaretz.co.il, @haaretz/htz-components

**Related issue(s):** #1886 (scruffy changed to santa


kedarguy:
removed santa from getHomePageElements, will add it to getSectionPageElements in a separate pull request where that file is created

PR #1971, opened 2019-06-10
fix(local storage): use try catch when setting items to local storage…

kedarguy:
…, and limit reading history to 50 items

affects: @haaretz/haaretz.co.il, @haaretz/htz-components


packages/components/htz-components/src/utils/criticalFontLoader/fontLoaderScript.js
```diff
@@ -13,10 +13,10 @@
 */
```
kedarguy: not using the util here because of the above comment

PR #1983, opened 2019-06-12
 feat(Mobile-Navigation): build a new mobile menu 

chenaviv:
**Related issue(s):** #0

## Description

GraphQl query changed to fit the data currently coming from dev backend (the added suggestedSections array)

Added canonicalUrl and a defult empty-string title to the finance app store, so that the initial mobileNavData query would not fail.

Added overlayMiscStyles property to A11yDialog, as well as a closingFunc property, that would be called instead of the default close function, when clicking the overlay or when the ESC key is pressed. 

Includes a commit that moves the reading-list initial query and the reading-list update to the graphql server.

**Notes:**

The pages for the most popular articles as well as for the reading list are not active in a non mobile device mode (desktop version have different URLs).
Also, the reading list page never seems to display any data (desktop version works fine)

## Checklist

  * [ ] Approved by designer

Usability:
  * [ ] Developer
  * [ ] Project manager

Tests
  
  * [ ] UI snapshot testing 
  * [ ] Unit testing
  * [ ] Integration testing
  * [ ] E2E testing


TxHawks:
Requires the `suggestedSections` field to come from PAPI. It's in pre right now, so this pull request can only be merged once papi in production is updated

PR #2031, opened 2019-06-27
fix(mobile_promoted_content): mobile promoted content

TawKhoury:

set gridContainer wrapper div display to grid. In TeaserResponsiveText set text if no mobileText

**Related issue(s):** #0

## Description

**Notes:**

## Checklist

  * [ ] Approved by designer

Usability:
  * [ ] Developer
  * [ ] Project manager

Tests
  
  * [ ] UI snapshot testing 
  * [ ] Unit testing
  * [ ] Integration testing
  * [ ] E2E testing


packages/components/htz-components/src/components/TeaserResponsiveText/TeaserResponsiveText.js
```diff
@@ -18,32 +18,35 @@ TeaserResponsiveText.defaultProps = {
...
 }: TextPropTypes): React.Node {
-  return (mobileText || text) && text !== mobileText ? (
+  const { css, theme, } = useFela();
+
+  // if there is no text or mobile text return nothing
+  if (!(mobileText || text)) return null;
+  // if there is text but no mobile text, and we are not in strict mode, return text for all breakpoints
+  // else, continue on
+  if (text && !mobileText && !displayStrict) {
+    return <React.Fragment>{text}</React.Fragment>;
```
TxHawks: Why the `<Fragment>` here? Any reason to not just return`text`?
```suggestion
    return text;
```

packages/components/htz-components/src/components/TeaserResponsiveText/TeaserResponsiveText.js
```diff
@@ -18,32 +18,35 @@ TeaserResponsiveText.defaultProps = {
...
+  }
+  // default situation, display mobile text on mobile breakpoints, and text on large breakpoints
```
TxHawks: This isn't the default - `displayStrict` is `false` by default.

packages/components/htz-components/src/components/TeaserResponsiveText/TeaserResponsiveText.js
TawKhoury: **After refactor**
When no React.Fragment. Flow error: 
Cannot return `text || mobileText` because:
- Either null or undefined [1] is incompatible with null [2].
- Or property `@@iterator` is missing in null or undefined [1] but exists in `$Iterable` [3]. 

(Iterable<?React$Node>)

Issue #2036, opened 2019-06-30
author components crash after registration


pelegash:
## Issue Type
**Bug fix**

## Description
author components crash after registration

**Assistive tech (screen readers):**

Before registration:
![image](https://user-images.githubusercontent.com/48085141/60396088-b39d7400-9b45-11e9-97a9-bab7ee3d7bf6.png)

After registration:
![image](https://user-images.githubusercontent.com/48085141/60396094-cadc6180-9b45-11e9-9ca6-ae72d8145a63.png)

PR #2041, opened 2019-07-01
fix(A11yDialog): prevent url changes resulting from non-modal dialogs

chenaviv:

**Related issue(s):** #2036

## Description

URL changes caused by certain `A11yDialog` instances resulted in the `closeDialog` method inside `AuthorAlert`'s `A11yDialog` to be invoked twice. Since that method sets the `returnFocusTo` element to null, the second invocation resulted in a typeError (cannot read 'focus' of null).

The added `isModal` condition in the `StatelessA11yDialog`'s `useEffect` that changes the current URL, fixes this particular issue.

Also added some basic safeguards in several pieces of code that call the focus method.

**Notes:**

## Checklist

  * [ ] Approved by designer

Usability:
  * [ ] Developer
  * [ ] Project manager

Tests
  
  * [ ] UI snapshot testing 
  * [ ] Unit testing
  * [ ] Integration testing
  * [ ] E2E testing


chenaviv:
Also added a useEffect that returns a function that sets the urlParamDialogstate to null, when component unmounts.

PR #2056, opened 2019-07-04
fix(live blog item): fix expanded item bug

TawKhoury:

**Related issue(s):** #0

## Description

**Notes:**

## Checklist

  * [ ] Approved by designer

Usability:
  * [ ] Developer
  * [ ] Project manager

Tests
  
  * [ ] UI snapshot testing 
  * [ ] Unit testing
  * [ ] Integration testing
  * [ ] E2E testing


packages/components/htz-components/src/components/ArticleTypes/LiveBlogArticle/LiveBlogElements/LiveBlogItem.js
```diff
@@ -31,290 +31,272 @@ const Fade = ({ children, fadeText, }) => (
...
-                </FelaComponent>
-                {/* Mobile action buttons end */}
-              </Grid>
-            </React.Fragment>
-          )}
-        </FelaTheme>
-      </React.Fragment>
-    );
-  }
+function useWindowWidth() {
```
TxHawks: A few things:
1) Since we aren't really interested in the size of the window, but rather whether it is within certain bounds, `useMedia` is a better fit here. It both communicates the intent better and is much cheaper, as events are only fired when actually needed.
2) `resize` events are fired extremely frequently, and sometimes unexpectedly, so a general purpose hook like that should definitely include a `throttle` attribute.
3) As I indicated in the first point, I don't think this is the correct use case for a `resize` event, but that doesn't mean it isn't actually a useful hook. Since it could be useful in many cases, it should live in the `hooks` folder, in a file of its own. [This is a good example for a general purpose resize hook](https://github.com/palmerhq/the-platform/blob/master/src/useWindowSize.tsx)

TawKhoury: Thanks. Done.

packages/components/htz-components/src/hooks/useScrollPosition.js
```diff
@@ -37,7 +37,7 @@ export default function useScrollPosition({
     return () => {
       window.removeEventListener('scroll', handle);
     };
-  }, []);
+  }, [ handle, ]);
```
TxHawks: The empty array here is intended. `throttle` can only be passed once, and `setScroll` is stable, so `handle` can be thought of as stable as well because the stale closure is enough. There is no need to detach and reattach the event listeners every time `handle` is recreated, which is on every render

packages/components/htz-components/src/hooks/useScrollXPosition.js
```diff
@@ -38,7 +38,7 @@ export default function useScrollXPosition({
     return () => {
       window.removeEventListener('scroll', handle);
     };
-  }, []);
+  }, [ handle, ]);
```
TxHawks: Same as above

packages/components/htz-components/src/hooks/useScrollYPosition.js
```diff
@@ -39,7 +39,7 @@ export default function useScrollYPosition({
     return () => {
       window.removeEventListener('scroll', handle);
     };
-  }, []);
+  }, [ handle, ]);
```
TxHawks: same as above

packages/components/htz-components/src/hooks/useWindowSize.js
```diff
@@ -0,0 +1,42 @@
+// @flow
+/* global window */
+import React from 'react';
+import throttle from '../utils/throttle';
+
+const events = new Set<() => void>();
+const onResize = () => events.forEach(fn => fn());
+
+export default function useWindowSize(options: { throttleMs?: number, } = {}) {
```
TxHawks:
```suggestion
export default function useWindowSize({ throttle: number, } = { throttle: 100, }) {
```

packages/components/htz-components/src/hooks/useWindowSize.js
```diff
@@ -0,0 +1,42 @@
...
+  const { throttleMs = 100, } = options;
```
TxHawks:
```suggestion
```

packages/components/htz-components/src/hooks/useWindowSize.js
```diff
@@ -0,0 +1,42 @@
...
+  const isServer = typeof window === 'undefined';
+
+  const [ size, setSize, ] = React.useState({
+    width: isServer ? 0 : window.innerWidth,
+    height: isServer ? 0 : window.innerHeight,
+  });
+
+  const handle = throttle(() => {
```
TxHawks:
```suggestion
  const handle = throttleFn(() => {
```

packages/components/htz-components/src/hooks/useWindowSize.js
```diff
@@ -0,0 +1,42 @@
+// @flow
+/* global window */
+import React from 'react';
+import throttle from '../utils/throttle';
```
TxHawks:
```suggestion
import throttleFn from '../utils/throttle';
```

packages/components/htz-components/src/hooks/useWindowSize.js
```diff
@@ -0,0 +1,42 @@
...
+    width: isServer ? 0 : window.innerWidth,
+    height: isServer ? 0 : window.innerHeight,
+  });
+
+  const handle = throttle(() => {
+    setSize({
+      width: window.innerWidth,
+      height: window.innerHeight,
+    });
+  }, throttleMs);
```
TxHawks:
```suggestion
  }, throttle);
```

packages/components/htz-components/src/hooks/useWindowSize.js
```diff
@@ -0,0 +1,42 @@
...
+    events.add(handle);
+
+    return () => {
+      events.delete(handle);
+
+      if (events.size === 0) {
+        window.removeEventListener('resize', onResize, true);
+      }
+    };
+  }, [ handle, ]);
```
TxHawks: Same as with `useScrollPosition`, no need to detach and reattach every time
```suggestion
  }, [ ]);
```

packages/components/htz-components/src/components/ArticleTypes/LiveBlogArticle/LiveBlogArticle.js
```diff
@@ -36,6 +37,10 @@ const IS_OSAKA_DISPLAYED = gql`
 
 function LiveBlog({ articleId, slots, path, isBlock, isExploded, site, }) {
   const getComponent = useGetComponent();
+
+  // responsive view when resizing window.
+  useWindowSize();
```
TxHawks: I still don't understand why use a `resize` listener and not a `matchMedia` listener. Wouldn't that be both more straightforward and more performant?

packages/components/htz-components/src/components/ArticleTypes/LiveBlogArticle/LiveBlogElements/LiveBlogItem.js
```diff
@@ -31,290 +32,260 @@ const Fade = ({ children, fadeText, }) => (
...
-                {/* Mobile action buttons end */}
-              </Grid>
-            </React.Fragment>
-          )}
-        </FelaTheme>
-      </React.Fragment>
-    );
-  }
+function LiveBlogItems(props) {
+  const mediaMatchS = useMedia({ query: { until: 's', }, });
```
TxHawks:
```suggestion
  const isBelowS = useMedia({ query: { until: 's', }, });
```

packages/components/htz-components/src/components/ArticleTypes/LiveBlogArticle/LiveBlogElements/LiveBlogItem.js
```diff
@@ -31,290 +32,260 @@ const Fade = ({ children, fadeText, }) => (
...
+  return <LiveBlogItem {...props} mediaMatchInMobile={mediaMatchS} />;
```
TxHawks:
```suggestion
  return <LiveBlogItem {...props} isBelowS={isBelowS} />;
```

PR #2110, opened 2019-07-23
fix(slug list): changes row gap from small to be the same as col gap

EnoshTs:

**Related issue(s):** #0

## Description

**Notes:**

## Checklist

  * [ ] Approved by designer

Usability:
  * [ ] Developer
  * [ ] Project manager

Tests
  
  * [ ] UI snapshot testing 
  * [ ] Unit testing
  * [ ] Integration testing
  * [ ] E2E testing


packages/components/htz-components/src/components/List/views/Slugs/SlugsView.js
```diff
@@ -95,7 +95,7 @@ export default function Slugs({
       areasTemplate={areasTemplate}
       innerBackgroundColor="transparent"
       rowTemplate="auto auto auto 1fr auto"
-      rowGap={[ { until: 's', value: '2rem', }, { from: 's', value: '1rem', }, ]}
+      rowGap={[ { until: 's', value: '2rem', }, { from: 's', value: '4rem', }, ]}
```
TxHawks: Since `rowGap` and `colGap` are now the same, we can drop them both and just use `gridGap`

PR #2154, opened 2019-08-01
fix(Gstat time scheduler): changed function to run every hour instead…

EnoshTs:
… of every day at 6 oclock added

affects: @haaretz/htz-components, @haaretz/graphql-server

**Related issue(s):** #0

## Description

**Notes:**

## Checklist

  * [ ] Approved by designer

Usability:
  * [ ] Developer
  * [ ] Project manager

Tests
  
  * [ ] UI snapshot testing 
  * [ ] Unit testing
  * [ ] Integration testing
  * [ ] E2E testing


packages/components/htz-components/src/utils/predicate.js
```diff
@@ -0,0 +1,64 @@
...
+
+  /***
+   * Validate variable empty
+   * Returns true if given variable is empty
+   * */
+  function empty(something) {
+
+    const _is = init(something);
+
+    const handleObject = object => Object.entries(object).length === 0;
```
TxHawks: `Object.keys` is actually sufficient here

packages/components/htz-components/src/utils/predicate.js
```diff
@@ -0,0 +1,64 @@
...
+  const _string = something => _type_of('string')(something);
+  const _number = something => _type_of('number')(something);
+  const _boolean = something => _type_of('boolean')(something);
+  const _array = something => Array.isArray(something);
+
+  /***
+   * Validate variable empty
+   * Returns true if given variable is empty
+   * */
+  function empty(something) {
```
TxHawks: What does empty mean? This is a confusing term is JS. is an empty string (`''`) empty? An empty array? Why is `false` considered empty? Why are negative numbers or a zero considered empty? You have to look in the code to actually understand what empty is. 

Also, quite a few utils are defined in the function scope, which are only used once. Their names don't help understand what is being check, so they also don't really help with readability.

I mean
```js
return _is(_null_or_undefined) ? true :
      _is(_object) ? handleObject(something) :
        (_is(_string) || _is(_array)) ? handleStringAndArray(something) :
          _is(_number) ? handleNumber(something) :
            _is(_boolean) ? handleBoolean(something) :
              true
```

When reading this, the `handleObject` doesn't help me understand what is being validated, so it actually obscures intent. `Object.keys(something)` is clearer re intent. But `objectIsEmpty(something)` would work too

packages/components/htz-components/src/utils/time.js
```diff
@@ -0,0 +1,31 @@
+import { not, } from './functional';
```
TxHawks: LIke I said earlier, time is really complicated. Let's just use `date-fns` which is tried and tested

PR #2210, opened 2019-08-19
fix(HtzLink): do not prevent default if link is believed to be opening in a new tab

chenaviv:

**Related issue(s):** #0

## Description

`ctrl+click` opens link in a new tab but keeps the focus on the current tab. `ctrl+shift+click` opens the link a new tab and moves the focus to that new tab (on my platform, at least). There is no sure way of replicating that behavior, so instead of using `window.open(href)`, default is not prevented when the ctrl key is pressed or when the platform is a desktop mac and the meta key is pressed.

**Notes:**

shift+click seems to open the link in a new window. Should that be addressed as well?

## Checklist

  * [ ] Approved by designer

Usability:
  * [ ] Developer
  * [ ] Project manager

Tests
  
  * [ ] UI snapshot testing 
  * [ ] Unit testing
  * [ ] Integration testing
  * [ ] E2E testing


TxHawks:
Yes, Let's not `preventDefault()` on `shift+click` too

TxHawks:
Also, `alt+click` saves the link's target, so I guess that should be excluded as well

PR #2223, opened 2019-08-21
Feature/service worker

kedarguy:


kedarguy:
add navigation preloaod

Issue #2172, opened 2019-08-05
Osaka & header fix


adibi83:
## Issue Type

**Update to existing feature**

## Description
1. Fix Osaka's hide mode when scrolling to the top of the page
2. Show header above Osaka on scroll up

PR #2231, opened 2019-08-25
feat(masthead/osaka): article masthead appears on upwards scroll/Osaka modified

chenaviv:

**Related issue(s):** #2172

## Description

Added a context provider for the masthead's 'shouldDisplay' boolean.
All dropdowns in menu use it to close the dropdown in the case that it is open and the Masthead's 'shouldDisplay' property is `false`.

The context defaults to `true`, so it does not effect the behavior of the dropdowns when they are rendered inside of the HompageMasthead.

Article masthead appears when scrolling up in views larger than mobile.
Scroll listener is not instantiated in mobile view.

Osaka is set to disappear at a higher Y position than before.

Added a `['masthead', -1]` z-index to top strip, as well as to the layoutRow for the preHeader,
so that the Osaka component would always render behind them.

**Notes:**

Not sure if this effects the behavior in Safari. I checked on one Mac and in BrowserStack, and it seemed to work fine.

## Checklist

  * [ ] Approved by designer

Usability:
  * [ ] Developer
  * [ ] Project manager

Tests
  
  * [ ] UI snapshot testing 
  * [ ] Unit testing
  * [ ] Integration testing
  * [ ] E2E testing


chenaviv:
Resolved conflict and rebased

PR #2363, opened 2019-10-16
Refactor/pages redundency

EranM6:
**Related issue(s):** #0

## Description

**Notes:**

## Checklist

  * [ ] Approved by designer

Usability:
  * [ ] Developer
  * [ ] Project manager

Tests
  
  * [ ] UI snapshot testing 
  * [ ] Unit testing
  * [ ] Integration testing
  * [ ] E2E testing


packages/apps/themarker.com/pages/article.js
```diff
@@ -0,0 +1,13 @@
+import React from 'react';
+
+import { Article, } from '@haaretz/htz-components';
+
+import getElements from '../utils/getArticlePageElements';
+
+
+export default props => (
```
TxHawks: Same as above re named function

packages/apps/haaretz.co.il/pages/article.js
```diff
@@ -0,0 +1,13 @@
...
+
+import getElements from '../utils/getArticlePageElements';
+
+
+export default props => (
+  <Article
+    {...props}
+    getElements={getElements}
+  />
+);
```
TxHawks: I'd rather this be a proper named function, so that it doesn't show up as `unnamed` in the dev tools:
```
export default function Article(props) {
  return (<Article getElements={getElements} {...props} />);
}
```

packages/components/htz-components/src/components/Article/Article.js
```diff
@@ -0,0 +1,362 @@
...
+    query: {
+      path,
+      isBlock,
+      isPreview,
+      isExploded,
+      articleType,
+    },
+  } = url;
+
+  const Component = getArticleComponent(articleType);
```
TxHawks: Maybe call this `ArticleTypeComponent` or something, so it's more obvious what it actually is in the dev tools

packages/components/htz-components/src/components/Article/Article.js
```diff
@@ -0,0 +1,362 @@
...
+import LayoutContainer from '../PageLayout/LayoutContainer';
+import ArticleSeo from '../ArticleSeo/ArticleSeo';
+
+const getArticleComponent = articleType => {
+  switch (articleType) {
+    case 'recipeArticle':
+      return dynamic(() => import('../ArticleTypes/RecipeArticle/RecipeArticle'));
+    case 'reviewArticle':
+      return dynamic(() => import('../ArticleTypes/ReviewArticle/ReviewArticle'));
+    case 'magazineArticle':
```
TxHawks: Wait, so you did include the magazine article here?

Since it's a different layout, should probably also be its own page. Will save that ternary around `<Component />` on line 276.

We can have an `<ArticleWrapper />` component to encapsulate the repeated parts between these two types of article layouts (basically everything except the ternary) and keep it dry.

PR #2386, opened 2019-10-24
feat(omnyPlayer): connect Omny custom player to their analytics API

chenaviv:
affects: @haaretz/htz-components, @haaretz/graphql-server

Adds Omny data check and handling of stored data in Article and in WidePageLayout

**Related issue(s):** #0

## Description

A single biAction for each Omny defined listening session is sent. This is intended for an evaluation of the discrepancies between the actual session count and the session count that is to be recorded at Omny. Note that Omny will save a session only if it had at least 10 seconds of playback, so some of the discrepancy may be a result of that.

A session is sent through the Omny API when:
  a. The audio playback has ended.
  b. There is an active session at the time of the HtzAudioPlayer unmounting.
  c. Found sessions that were saved to localStorage when running the OmnyDataCheck component.

A session is saved to localStorage when:
  a. The posting of data has failed in cases 'a' or 'b' from the above section.
  b. There is an active session at the time of a 'beforeUnload' or 'pagehide' event on the window.

Currently not checking the response from Omny to verify a successful POST request,
but treating any request that doesn't error out as successful.

**Notes:**

## Checklist

  * [ ] Approved by designer

Usability:
  * [ ] Developer
  * [ ] Project manager

Tests
  
  * [ ] UI snapshot testing 
  * [ ] Unit testing
  * [ ] Integration testing
  * [ ] E2E testing


chenaviv:
Added a biAction call for most audio player actions.

PR #2512, opened 2019-11-20
feat(chart beat): set up charbeat config hostname according to site

kedarguy:


kedarguy:
@TxHawks did you close without merging on purpose?

